PR #5, opened 2016-05-20
Point contextkit to the mer-core-attic

theappleman:
Git repository changed upstream:

ERROR: Fetcher failure: Fetch command failed with exit code 128, output:
Cloning into bare repository '/home/asteroid/workspace/asteroid-image-bass/build/downloads/git2/git.merproject.org.mer-core.nemo-qml-plugin-contextkit.git'...
fatal: could not read Username for 'https://git.merproject.org': No such device or address

ERROR: Function failed: Fetcher failure for URL: 'git://git.merproject.org/mer-core/nemo-qml-plugin-contextkit.git;protocol=https'. Unable to fetch URL from any source.


FlorentRevest:
Thank you!

PR #10, opened 2016-09-05
Enable SELinux bootparam

theappleman:
Necessary to disable SELinux from the kernel command line

  selinux=0


FlorentRevest:
https://github.com/AsteroidOS/meta-asteroid/blob/master/scripts/check-config#L219 I'm surprised that this is not enough? but if you can confirm that even without those config options you still get selinux in your kernel and you need to disable it with bootparam then that's fine, we can merge your PR.

theappleman:
This patch is not necessary and can be closed.

PR #12, opened 2017-01-11
Fix SDK build errors.

aebruno:
This fixes AsteroidOS/asteroid#10. 

bitbake meta-toolchain-qt5 was failing with the following error when compiling qtwebkit:

```
asteroid/build/tmp-glibc/sysroots/bass/usr/include/c++/6.2.0/cstdlib:75:25:
fatal error: stdlib.h: No such file or directory
```

This appears to be related to how qmake passes some include paths with "-isystem" prefix [1,2]. This flag effects the internal header search path order and seems to cause issues when compiling qtwebkit for the sdk. By default QMAKE_CFLAGS_ISYSTEM = -isystem see [3]. This patch resets QMAKE_CFLAGS_ISYSTEM to use -I instead for qtwebkit.

Additionally, the packagegroup-qt5-toolchain-target.bbappend adds/removes some packages that were causing the do_populate_sdk to fail. These changes were described in more detail here [4].

[1] http://lists.openembedded.org/pipermail/openembedded-core/2016-August/125952.html
[2] http://stackoverflow.com/questions/37218953/isystem-on-a-system-include-directory-causes-errors
[3] http://code.qt.io/cgit/qt/qtbase.git/tree/mkspecs/common/gcc-base.conf?h=5.6#n47
[4] http://ronan.lepage1.free.fr/gumstix-wiki/doku.php?id=asteroidos:clean_start


FlorentRevest:
This looks really good to me. Thanks a lot!

PR #13, opened 2017-02-18
Remove absolute dir from build definition

daniellandau:
Before this change I got a failure:

```
| /home/dlan/local/src/asteroid/build/tmp-glibc/sysroots/x86_64-linux/usr/libexec/arm-oe-linux-gnueabi/gcc/arm-oe-linux-gnueabi/6.2.0/ld: warning: library search path "/usr/lib" is unsafe fo
r cross-compilation
| /usr/lib/libQt5Network.so: file not recognized: File format not recognized
```


FlorentRevest:
Alright, I merge this but it should soon be replaced by applying a .patch

PR #15, opened 2017-05-30
[RFC] Composite USB device

atx:
This PR removes usb-moded in favor of a single-configuration USB composite device providing  ADB, CDC ECM and MTP at once.

### DIFF
--- a/recipes-asteroid/asteroid-usbsetup/asteroid-usbsetup/asteroid-usbsetup.sh
+++ b/recipes-asteroid/asteroid-usbsetup/asteroid-usbsetup/asteroid-usbsetup.sh
@@ -1,0 +1,38 @@
+#! /bin/sh
+
+set -e
+
+ANDROID_USB="/sys/class/android_usb/android0"
+
+log_info () {
+	echo "[INFO] $@"
+}
+
+start_usb () {
+	echo "0" > $ANDROID_USB/enable
+
+	log_info "Starting adb mode"
+
+	manufacturer="$(cat /system/build.prop | grep -o 'ro.product.manufacturer=.*' | cut -d'=' -f 2)"
+	model="$(cat /system/build.prop | grep -o 'ro.product.model=.*' | cut -d'=' -f 2)"
+	serial="$(cat /proc/cmdline | sed 's/.*androidboot.serialno=//' | sed 's/ .*//')"
+
+	echo $manufacturer > $ANDROID_USB/iManufacturer
+	echo $model        > $ANDROID_USB/iProduct
+	echo $serial       > $ANDROID_USB/iSerial
+
+	echo 18d1 > $ANDROID_USB/idVendor
+	echo 0a03 > $ANDROID_USB/idProduct
+
+	mkdir -p /dev/usb-ffs/adb
+	mount -t functionfs adb /dev/usb-ffs/adb
+
+	echo adb > $ANDROID_USB/f_ffs/aliases
+	echo ffs,ecm,mtp > $ANDROID_USB/functions
+
+	echo "1" > $ANDROID_USB/enable
+
+	log_info "Initialized android_usb with $manufacturer $model $serial"
+}
+
+start_usb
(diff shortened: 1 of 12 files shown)


recipes-asteroid/asteroid-usbsetup/asteroid-usbsetup/asteroid-usbsetup.sh
```diff
@@ -0,0 +1,38 @@
...
+ANDROID_USB="/sys/class/android_usb/android0"
+
+log_info () {
+	echo "[INFO] $@"
+}
+
+start_usb () {
+	echo "0" > $ANDROID_USB/enable
+
+	log_info "Starting adb mode"
```
FlorentRevest: Why indicating "adb mode" when you actually activate ffs ecm and mtp?

recipes-ubuntu/mtp-server/mtp-server/mtp-server.service
```diff
@@ -0,0 +1,14 @@
+[Unit]
+Description=mtp-server service
+Requires=usb-setup.service dev-mtp_usb.device
```
FlorentRevest: Haven't you forgot "After=usb-setup.service" ?

FlorentRevest (changes requested):
I do agree with the main principles of your changes, and ecm works surprisingly well, congratulations and thank you for your work!

However, there are a few things I would want to see changed before allowing your patches to be merged:
- Some of them are described in code comments  attached to this review
- An error happens in the kernel when trying to start mtp-server. https://gist.github.com/FlorentRevest/f0350240e9d44f10bab7088cef4078ef are you aware of it?
- Every ~10 seconds USB connection is interrupted, is it a known problem?
- dev-mtp-usb.device timeouts according to https://gist.github.com/FlorentRevest/b468f9b308b1e449f5440ff1e31b41ed I haven't been able to test MTP.
- Is connman supposed to be kept installed? If you plan to remove it, have you thought about packages depending on it?
- DSME keeps track of USB State, could you check whether it's important to its inner working and if we can ignore it or replace the usage of usb-moded to your custom solution.
- lipstick also relies on usb-moded to show handy notifications when connected or disconnected. Have you thought of a solution to keep lipstick informed of this kind of events?
- asteroid-settings currently has a "USB page" relying heavily on usb-moded. If we move toward a composite device exposing almost every capabilities at once, I would want to replace the USB page with a Developer Mode page showing the IP address of the watch and maybe also a randomly generated ceres password (just like in SailfishOS). I would also want this Developer Mode to be enabled or disabled at will, allowing to turn adb and ecm off. If the overall state of your solution gets improved I may be able to help you on this task.

Keep up the good work!

atalax:
> An error happens in the kernel when trying to start mtp-server. https://gist.github.com/FlorentRevest/f0350240e9d44f10bab7088cef4078ef are you aware of it?

Yep, happens on lenok too, I didn't notice that. That's probably worthy of debugging.

> Every ~10 seconds USB connection is interrupted, is it a known problem?

This is likely caused by a ["feature"](https://sourceforge.net/p/libmtp/code/ci/master/tree/src/libusb1-glue.c#l1973) in libmtp combined with gvfsd automounting the device (and timeouting as you are missing /dev/mtp_usb for some reason). The proper way to fix that is getting a product ID and sending a patch to libmtp. The less-than-proper way is to steal a vid/pid which is handled correctly in libmtp.

> dev-mtp-usb.device timeouts...

I am not really sure about why is that the case, dory kernel has f_mtp support so udev should just spawn /dev/mtp_usb after mtp gets activated.

> Is connman supposed to be kept installed? If you plan to remove it, have you thought about packages depending on it?

It's not interfering with anything so far. Which packages depend on it? I have a build without it installed (for experimenting with systemd-resolved) and everything seems to work fine (is it enough just as a build dependency?)

> DSME keeps track of USB State, could you check whether it's important to its inner working and if we can ignore it or replace the usage of usb-moded to your custom solution.

I am not sure. Looking through the sources, it looks like DSME might end up in an "actdead" state instead of shutting down if it does not know charger state. I don't really know if that is a problem though.

> lipstick also relies on usb-moded to show handy notifications when connected or disconnected. Have you thought of a solution to keep lipstick informed of this kind of events?

I completely missed that. I am going to add an udev rule for handling this.

> asteroid-settings currently has a "USB page" relying heavily on usb-moded.

I am aware of that. I wanted to get this finalized before starting to patch the interface.

FlorentRevest:
Closing as stale for many years

PR #17, opened 2017-11-08
Fix issue with adbd patch striplevel

theappleman:
0001-Allow-adbd-to-be-ran-as-root.patch contains paths with 'a/', 'b/' prefixes which need to be stripped.

Buildlog snippet:
https://gist.github.com/theappleman/1554e48bc17d32cf88faae3eb94587cd


FlorentRevest:
I'm surprised I haven't met this problem before (neither on my machine nor on the nightlies build server) but apprently, applying your patch causes no regression either so I'll merge the PR.

Thank you AppleMan!

PR #27, opened 2020-01-25
Add support for Bluetooth HID and Audio

MagneFire:
This PR includes the needed changes to allow for Bluetooth HID and Audio to work.

A seperate adjustment is needed to patchram.service to fix extreme skipping. That adjustment consists of setting a baudrate. I will make a followup PR for sturgeon with this fix.

You may want to change ControllerMode = le to dual, in main.conf, if a Bluetooth peripheral is not visible when scanning. This is not included in this PR as that change may impact battery life.
We may document this on the asteroidos.org wiki or I think it will be cool if we can change the controller mode in asteroid-settings. Maybe even add the ability to connect Bluetooth peripherals via asteroid-settings :)

### DIFF
--- a/recipes-connectivity/bluez5/bluez5/input.conf
+++ b/recipes-connectivity/bluez5/bluez5/input.conf
@@ -1,0 +1,13 @@
+# Configuration file for the input service
+
+# This section contains options which are not specific to any
+# particular interface
+[General]
+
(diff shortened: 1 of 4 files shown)


FlorentRevest:
Great job, as usual :) Thanks a lot @MagneFire

Issue #16, opened 2017-10-11
android-simg rootfs format


FlorentRevest:
Currently, images generated by the asteroid build system only take a minimal amount of space and don't fill the entire userdata partition.
This has caused several kinds of problems (like a partition getting filled quickly and preventing the system from working correctly) and generating a larger rootfs partitions should provide a better experience.
However, having large partitions may fill the release.asteroidos.org server and cause long download times. It turns out that fastboot support an "simg" file format which addresses precisely this problem. Defining a new sparse partition bbclass for OE using a tool like img2simg would be super useful.

beschulz:
would it be an option to use img2simg to convert the images into sparse images?
I used

resize2fs 2800M asteroid-image-dory.ext4
img2sim asteroid-image-dory.ext4 asteroid-image-dory.ext4.sparse
fastboot format:ext4 userdata
fastboot flash userdata asteroid-image-dory.ext4.sparse

it appears to be working fine and the sparse image is only 357M

FlorentRevest:
@beschulz For a detailed description of why we don't distribute simgs, I'd point you to that comment: https://github.com/AsteroidOS/asteroid/issues/31#issuecomment-348731212
Essentially, we can't get the "temporary installation" to work with simgs and I think this is very appreciated by testers who don't want to remove WearOS.
Also, we don't have data on maximum size of userdata partitions on each supported watch.

comradekingu:
Isn't this something http://atterer.org/jigdo/ solves?

FlorentRevest:
@comradekingu  Mhh, I don't think asking our users to download a downloader is really appropriate. I'd rather ask them to install img2simg and resize2fs.
What I think could be a good solution would be to actually resize the rootfs to fill the userdata partition from the initrd on the first boot of a real installation. This way, we don't have to redistribute both a simg and an img, we have a flexible solution that doesn't require to know the partition size of each device at compile time and we don't have to use tools such as resize2fs or img2simg.

PR #28, opened 2020-01-31
resize-rootfs: Expand rootfs when installed on userdata.

MagneFire:
This PR adds a script that uses resize2fs to resize the rootfs to fill the userdata partition when installed on userdata.
When this script is executed on a temporary installation it will exit without using resize2fs.

This PR possibly implements #16.

Please let me know if this PR fulfills the quality requirements to be merged with meta-asteroid.

Couple of uncertainties from my end:
- Is the script located at the proper location(/usr/bin) or should I move this to a reserved location(/var)
- This PR does not incorporate a reboot mechanism. I noticed that Raspbian and Armbian warn that the system should be rebooted when on-line resize is performed.

### DIFF
--- a/classes/asteroid-image.bbclass
+++ b/classes/asteroid-image.bbclass
@@ -5,7 +5,7 @@ LICENSE = "GPL-2.0"
 IMAGE_FEATURES += "package-management debug-tweaks"
 
 IMAGE_INSTALL += " \
-base-files base-passwd systemd busybox iproute2 connman pam-plugin-loginuid bluez5 pulseaudio-server openssh-sshd openssh-sftp-server openssh-scp statefs dsme mce ngfd timed sensorfw android-init mapplauncherd-booster-qtcomponents usb-moded ofono \
+base-files base-passwd systemd busybox iproute2 connman pam-plugin-loginuid bluez5 pulseaudio-server openssh-sshd openssh-sftp-server openssh-scp statefs dsme mce ngfd timed sensorfw android-init resize-rootfs mapplauncherd-booster-qtcomponents usb-moded ofono \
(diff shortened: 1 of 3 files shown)


FlorentRevest:
Thanks so much for looking into this, this is awesome!

Have you considered having this script start only once on the first boot instead of everytime the user boots...? There is an easy way to do this with OpenEmbedded https://github.com/AsteroidOS/meta-asteroid/blob/master/recipes-nemomobile/statefs/statefs_git.bb#L51 However... please note that this mechanism seems to be randomly broken for whatever reason (cf: https://github.com/AsteroidOS/asteroid/issues/94 and https://github.com/AsteroidOS/asteroid/issues/95 ) if you figure out why and fix both things I will arrange a cookies delivery to an address of your convenience (I actually mean it).

Have you considered shrinking the extra free space we statically add to every asteroidos rootfs when we generate them https://github.com/AsteroidOS/meta-asteroid/blob/master/classes/asteroid-image.bbclass#L18 ? (this can be done in a different commit or a different PR, whatever you like most)

I'm also surprised you can resize the partition on the running system, I thought you had to do that from the initramdisk before the partition was mounted. But if you have been able to run this without any issue, then it is probably fine.

This would fix #16.

MagneFire:
I did think about running this only on the first boot, but I didn't know about pkg_postinst_ontarget_${PN}. And resize2fs will only grow the partition when it can. So if it is already grown to the partition size it won't do anything. In this case it might slow down the boot time a little bit... I will look into the issues. If I can fix that then the pkg_postinst_ontarget_${PN} would be the best solution.

I will also look into reducing the extra free space.

FlorentRevest:
This was reported by the latest nightly build https://pastebin.com/RRp3iDHg

MagneFire:
Oops, I was missing the SRC_URI.
I tried to put the script contents in the pkg_postinst_ontarget section,  but this results into an empty package. This is ignored by opkg...

PR #29, opened 2020-02-06
mce & sensorfw use latest revision.

MagneFire:
Use the latest revision of mce and sensorfw as they are under the AsteroidOS organization.


FlorentRevest:
Doesn't the mce repository require a PR that applies that patches that got removed from here?

MagneFire:
No, @lpotter already applied those patches before working on his implementation of the wrist gesture sensor. I just finished that.

FlorentRevest:
All good then ;) I missed that part.

PR #30, opened 2020-02-10
resize-rootfs: Move script contents to recipe.

MagneFire:
Once again, sorry for not testing this properly...
 I have now moved the script contents to the recipe. It previously didn't seem to run the script. But with ALLOW_EMPTY_${PN} = "1" this seems to be solved.


FlorentRevest:
Looks good, thanks a lot :)
No need to be sorry, I should have reviewed it more carefully as well ;) As always, thanks a lot for your time and contributions, really appreciated.

PR #31, opened 2020-02-15
openssh: fix login on old kernels with openssl 1.1.1d

MagneFire:
It looks like OpenSSL was updated to 1.1.1d from oe-core (https://github.com/openembedded/openembedded-core/commits/warrior). This appears to break OpenSSH with older kernels.
A fix has already been made upstream. This should be included in 8.2 release. I think it will take some time before we get this release in the warrior branch.

Instead of using this patch, we may also freeze the oe-core (https://github.com/AsteroidOS/asteroid/blob/master/prepare-build.sh#L74) to a specific SRCREV. This makes everything less susceptible to bugs in the oe-core branch at the cost of having outdated packages.

This patch will also need to be removed when we inevitably get the OpenSSH 8.2 release in the oe-core repo.

This also fixes https://github.com/AsteroidOS/meta-mtk6580-hybris/issues/8

References:
https://github.com/z3ntu/aports/commit/ccc44317fa09f12c982b5d91aba374e81725bea6
https://github.com/openssl/openssl/issues/9984
https://github.com/openssh/openssh-portable/pull/149


FlorentRevest:
Looks good, thank you ;)

PR #32, opened 2020-02-15
libsdl2: fix compilation and runtime errors.

MagneFire:
OpenGL is enabled in DISTRO_FEATURES. But opengl isn't actually provided by libhybris.
We also need an upstream patch to fix runtime errors with older wayland protocol specs.

This fixes the following runtime issue:
wl_registry@2: error 0: invalid version for global wl_data_device_manager (4): have 1, wanted 3


FlorentRevest:
Fair enough, we don't really support any SDL app by default but it can't hurt either.
Thanks! :D

PR #33, opened 2020-02-24
ngfd-plugin-droid-vibrator: Rename alarm to clock.

MagneFire:
The ffmemless version uses the clock pattern for alarm events.
This caused alarm.ini to refer to effect = clock. Which does not exist for the non-ffmemless version.
Renaming the alarm event to clock fixes this issue.

This should fix https://github.com/AsteroidOS/meta-wren-hybris/issues/3 and fix https://github.com/AsteroidOS/meta-lenok-hybris/issues/6

### DIFF
--- a/recipes-nemomobile/ngfd/ngfd-plugin-droid-vibrator/50-droid-vibrator.ini
+++ b/recipes-nemomobile/ngfd/ngfd-plugin-droid-vibrator/50-droid-vibrator.ini
@@ -1,7 +1,7 @@
 [droid-vibrator]
 
 # For each string in EFFECT_LIST define a sequence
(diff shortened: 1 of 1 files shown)


FlorentRevest:
Awesome, good catch! :)
To be fair, ffmemless is of questionable interest by now. Back when I started the project, the Android ngfd backend had all sorts of limitations such as not being able to sustain a vibration for a configurable number of seconds. But now it can be tweaked just as much as ffmemless and without requiring a hand crafted kernel patch. It would probably be wise to drop the ffmemless backend and patches at some point in the future to avoid that sort of maintenance pain... :)
That being said, thanks a lot for your debugging, much appreciated ;)

PR #36, opened 2020-03-12
asteroid-image: package-index depends on opkg-make-index.

MagneFire:
In a previous commit I provided a fix for the repository not being generated. However, it seems that I forgot to include the dependency for the `opkg-make-index` tool.
This results in errors during a clean build.

This fixes https://github.com/AsteroidOS/asteroid/issues/115


FlorentRevest:
Looks good, thank you :)

PR #38, opened 2020-05-15
bumps SRCREV

jrtberlin:


MagneFire:
Why not change this to `${AUTOREV}`?
The `asteroid-wallpapers` repository is already under the AsteroidOS umbrella.

jrtberlin:
@MagneFire I have not done this bc it was not set before. Maybe it was intentional. For assets like wallpapers `${AUTOREV}` is probably the saner option.

FlorentRevest:
Yeah, in theory this repo don't change often and a static commit hash doesn't require too much work. But still `${AUTOREV}` would require less work and since all asteroid-* repos have AUTOREV I think that would be a better option here.

PR #40, opened 2020-05-29
linux: Add kernel modules to image.

MagneFire:
Some ports require kernel modules to be packaged with rootfs.
This patch fixes sensors not working on sparrow.
Big thanks to @hummlbach for providing help in fixing this issue!

This patch was originally applied to the meta-sparrow platform only. This seems general enough to be included for all platforms. Other platforms swift, anthias and tetra also use kernel modules which likely aren't included in the final image. This may fixes issues on those platforms too, though not tested!

Close this PR if it turns out that this PR has negative effects on other platforms.

### DIFF
--- a/classes/asteroid-image.bbclass
+++ b/classes/asteroid-image.bbclass
@@ -5,7 +5,7 @@ LICENSE = "GPL-2.0"
 IMAGE_FEATURES += "package-management debug-tweaks"
 
 IMAGE_INSTALL += " \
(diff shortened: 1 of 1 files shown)


FlorentRevest:
argh you were faster than I could be!
I was about to tell you on the other PRs that this should be a meta-asteroid change :) great job

FlorentRevest:
may I ask that you update the asteroidos.org compatibility matrices for watches that have support improvements? mainly sparrow for now I guess

MagneFire:
I guess tilt-to-wake is the main thing to add, but the matrix falls of the screen on mobile...
compass hasn't changes, because it looks like this isn't provided by the backend.

FlorentRevest:
sure, still worth adding

PR #46, opened 2020-09-01
qtdbusextended: Fix setting mpris properties.

MagneFire:
When setting the volume, shuffle, loop status it would fail with:
error time=1598646695.682396 sender=:1.3 -> destination=:1.12 error_name=org.freedesktop.DBus.Error.UnknownMethod reply_serial=69
    string "No such method 'Set' in interface 'org.freedesktop.DBus.Properties' at object path '/org/mpris/MediaPlayer2' (signature 'ssd')"

This is due to qtdbusextended setting the properties using the type and value immediately (double:0.4), while the format variant:type:value(varient:double:0.4) is expected.

This commit fixes this issue by wrapping the type and value in a variant.

It took me quite some time to figure out what was going wrong :cry: 

Some useful things that may be helpful for others are:
- Use dbus-monitor, to look at all dbus messages.
- Use the `dbus-send --print-reply --dest=org.mpris.MediaPlayer2.asteroid-btsyncd /org/mpris/MediaPlayer2 org.freedesktop.DBus.Properties.Set string:org.mpris.MediaPlayer2.Player string:Volume variant:double:0.4` command to set the volume from the console.
- `dbus-send --print-reply --dest=org.mpris.MediaPlayer2.asteroid-btsyncd /org/mpris/MediaPlayer2 org.freedesktop.DBus.Properties.Get string:org.mpris.MediaPlayer2.Player string:Volume` Gets the current volume.
- `dbus-send --print-reply --dest=org.mpris.MediaPlayer2.asteroid-btsyncd /org/mpris/MediaPlayer2 org.mpris.MediaPlayer2.Player.PlayPause` play/pause playback.
- `dbus-send --print-reply --dest=org.mpris.MediaPlayer2.asteroid-btsyncd /org/mpris/MediaPlayer2 org.freedesktop.DBus.Introspectable.Introspect` Get list of all methods/properties.

### DIFF
--- a/recipes-nemomobile/qtdbusextended/qtdbusextended/0001-Use-variant-when-setting-properties.patch
+++ b/recipes-nemomobile/qtdbusextended/qtdbusextended/0001-Use-variant-when-setting-properties.patch
@@ -1,0 +1,30 @@
+From 78231f00b786e8da20a49da2861285ae440b0bf8 Mon Sep 17 00:00:00 2001
+From: MagneFire <IDaNLContact@gmail.com>
+Date: Mon, 31 Aug 2020 19:03:58 +0200
+Subject: [PATCH] Use variant when setting properties.
+ org.freedesktop.DBus.Properties.Set expects a variant containing a type and
+ value(variant:double:0.4). Previously it would just send the type and value
+ (double:0.4). This would result in the following error:   string "No such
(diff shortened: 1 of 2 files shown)


FlorentRevest:
Woohoo great job thanks! :)

PR #47, opened 2020-09-12
qtdbusextended: Fix recipe parse error.

MagneFire:
Oops, somehow it was working on my machine.

Anyway, adding a backslash fixes the recipe parse error.


FlorentRevest:
Oh, those are annoying :) Thanks for spotting and fixing that

PR #50, opened 2020-10-06
qml-asteroid: Use a configuration file for device specific settings.

MagneFire:
Adjust recipe to use a `machine.conf` file when it is provided by a meta-*-hybris layer. Otherwise the defaults are assumed.

### DIFF
--- a/recipes-asteroid/qml-asteroid/qml-asteroid_git.bb
+++ b/recipes-asteroid/qml-asteroid/qml-asteroid_git.bb
@@ -11,33 +11,8 @@ PV = "+git${SRCPV}"
 S = "${WORKDIR}/git"
 inherit qmake5
 
-PACKAGE_ARCH = "${MACHINE_ARCH}"
-
-DEPENDS += "qtdeclarative qtsvg qtvirtualkeyboard mlite mapplauncherd-booster-qtcomponents qtdeclarative-native"
-RDEPENDS_${PN} += "qtsvg-plugins qtvirtualkeyboard asteroid-icons-ion"
-
-do_configure_prepend() {
-    if [ ${MACHINE_DISPLAY_FLAT_TIRE} ]
-    then
-        export EXTRA_QMAKEVARS_PRE="${EXTRA_QMAKEVARS_PRE} DEFINES+=FLAT_TIRE=${MACHINE_DISPLAY_FLAT_TIRE}"
-    fi
-    if [ ${MACHINE_DISPLAY_BORDER_GESTURE_WIDTH} ]
-    then
-        export EXTRA_QMAKEVARS_PRE="${EXTRA_QMAKEVARS_PRE} DEFINES+=BORDER_GESTURE_WIDTH=${MACHINE_DISPLAY_BORDER_GESTURE_WIDTH}"
-    fi
-    if [ ${MACHINE_DISPLAY_ROUND} = "true" ]
-    then
-        export EXTRA_QMAKEVARS_PRE="${EXTRA_QMAKEVARS_PRE} DEFINES+=ROUND_SCREEN"
-    fi
-    if [ ${MACHINE_HAS_WLAN} = "true" ]
-    then
-        export EXTRA_QMAKEVARS_PRE="${EXTRA_QMAKEVARS_PRE} DEFINES+=HAS_WLAN"
-    fi
-    if [ ${MACHINE_HAS_SPEAKER} = "true" ]
-    then
-        export EXTRA_QMAKEVARS_PRE="${EXTRA_QMAKEVARS_PRE} DEFINES+=HAS_SPEAKER"
-    fi
-}
+DEPENDS += "asteroid-machine-config qtdeclarative qtsvg qtvirtualkeyboard mlite mapplauncherd-booster-qtcomponents qtdeclarative-native"
+RDEPENDS_${PN} += "asteroid-machine-config qtsvg-plugins qtvirtualkeyboard asteroid-icons-ion"
 
 FILES_${PN} += "/usr/lib /usr/share/icons/asteroid/"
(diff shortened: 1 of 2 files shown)


recipes-asteroid/qml-asteroid/qml-asteroid_git.bb
```diff
@@ -11,34 +11,17 @@ PV = "+git${SRCPV}"
...
-    then
-        export EXTRA_QMAKEVARS_PRE="${EXTRA_QMAKEVARS_PRE} DEFINES+=HAS_WLAN"
-    fi
-    if [ ${MACHINE_HAS_SPEAKER} = "true" ]
-    then
-        export EXTRA_QMAKEVARS_PRE="${EXTRA_QMAKEVARS_PRE} DEFINES+=HAS_SPEAKER"
+do_install_append() {
+    if [ -f ../machine.conf ]; then
+      install -d ${D}/etc/
+      cp ../machine.conf ${D}/etc/machine.conf
```
FlorentRevest, line 20: As discussed in https://github.com/AsteroidOS/meta-sturgeon-hybris/pull/13/files#r500633781 I would prefer if we'd dynamically generate this file.

recipes-asteroid/qml-asteroid/qml-asteroid_git.bb
```diff
@@ -11,34 +11,17 @@ PV = "+git${SRCPV}"
 S = "${WORKDIR}/git"
 inherit qmake5
 
-PACKAGE_ARCH = "${MACHINE_ARCH}"
```
FlorentRevest, line 14: But this is still machine specific :thinking:

recipes-asteroid/qml-asteroid/qml-asteroid_git.bb
```diff
@@ -11,34 +11,17 @@ PV = "+git${SRCPV}"
...
+do_install_append() {
+    if [ -f ../machine.conf ]; then
+      install -d ${D}/etc/
+      cp ../machine.conf ${D}/etc/machine.conf
     fi
 }
 
-FILES_${PN} += "/usr/lib /usr/share/icons/asteroid/"
+
+FILES_${PN} += "/etc/ /usr/lib /usr/share/icons/asteroid/"
```
FlorentRevest, line 25: I think the configuration file should be distributed as a separate opkg. It makes sense for qml-asteroid to be one package that can be installed on any watch but not the config file.

recipes-asteroid/qml-asteroid/qml-asteroid_git.bb
FlorentRevest: I can't remember whether there is a way to specify that only a subpackage, say, qml-asteroid-config, is machine specific.
Maybe you could also create a separate recipe for the machine config because even though qml-asteroid will be its main (only) user for now, it makes sense for other components to look into this file. And that recipe would have PACKAGE_ARCH = "${MACHINE_ARCH}"

FlorentRevest (changes requested):

MagneFire:
Thanks for the feedback!
All your requested changes should be solved by moving the machine specific stuff to a separate recipe :smile:

recipes-asteroid/qml-asteroid-config/qml-asteroid-config_1.0.bb
```diff
@@ -0,0 +1,31 @@
...
+    for v in machine_vars:
+        if d.getVar(v) is not None:
+            config.set('General', v[len("MACHINE_"):], str(d.getVar(v)))
+
+    with open(d.getVar('WORKDIR') + "/machine.conf", "w") as machine_conf:
+        config.write(machine_conf)
+}
+
+do_install_append() {
+    if [ -f ../machine.conf ]; then
```
FlorentRevest, line 23: Do we want to ignore if this file is missing?
My understanding is that your do_generate_config would always write a file. If the file is missing, this is probably worth breaking the build here instead of silently continuing without a machine config.

recipes-asteroid/qml-asteroid-config/qml-asteroid-config_1.0.bb
```diff
@@ -0,0 +1,31 @@
...
+      install -d ${D}/etc/qml-asteroid/
+      cp ../machine.conf ${D}/etc/qml-asteroid/machine.conf
```
FlorentRevest, line 25: In principle we should use install to install all files (especially system configuration files) and explicitly set permissions.
In practice... I know it's easier to just use cp and I've done that everywhere, but I regret. :p

recipes-asteroid/qml-asteroid-config/qml-asteroid-config_1.0.bb
```diff
@@ -0,0 +1,31 @@
...
+    fi
+}
+
+addtask do_generate_config before do_install
+
+FILES_${PN} += "/etc/qml-asteroid/"
```
FlorentRevest, line 31: I wonder whether it makes sense to have this under the qml-asteroid umbrella. In principle, any Asteroid component could be interested in this config file, for example if you have a shell script which does one thing or the other depending on whether the machine has MACHINE_HAS_SPEAKER or not, it would be absurd to have it depend on a qml-asteroid config. What do you think of elevating this as a global asteroid config file?

FlorentRevest (commented):
Great job :) just a few nits

recipes-asteroid/qml-asteroid-config/qml-asteroid-config_1.0.bb
MagneFire: My initial reason for having this was that `qml-asteroid` doesn't crash when the config file doesn't exist. But you are correct, that `do_generate_config` always creates the `machine.conf`.

I will adjust this so that the build will fail when the `do_generate_config` task fails.

recipes-asteroid/qml-asteroid-config/qml-asteroid-config_1.0.bb
MagneFire: Excellent point, will change to /etc/asteroid

PR #55, opened 2020-12-18
Add and enable Color Emoji.

MagneFire:
This PR adds the necessary stuff to make color emoji work.
Some of these commits can be adjusted/removed when we eventually upgrade to something more recent than `warrior`:
`freetype: Enable pixmap`: Is enabled here https://git.yoctoproject.org/cgit/cgit.cgi/poky/commit/meta/recipes-graphics/freetype?id=2673d8005c1c6ed27a8a7e3903b904e0c0711320
`ttf-noto-emoji: Add Google Noto Emoji`: Added here: https://github.com/openembedded/meta-openembedded/commit/08a97cb0e03b7b14bf4ef2632f2db24a86ccae8b

I tried to make the `69-emoji.conf` work system wide, but could not get it to work for some reason, what I tried:
- Copy `69-emoji.conf` to /usr/share/fontconfig/conf.avail/
- Copy `69-emoji.conf` to /etc/fonts/conf.d/
- Change the binding from `weak` to `strong`
- Rename `69-emoji.conf` to `00-emoji.conf` and `99-emoji.conf`
- Run `fc-cache -f -v` on every change to refresh the font cache (run for both root and ceres).

Depending on what you prefer, we could also merge this PR into https://github.com/AsteroidOS/asteroid-fonts.

Preview:
Screenshot_20201216_145722

Screenshot_20201216_145733

PR #56, opened 2020-12-19
Add and enable Color Emoji. (v2)

MagneFire:
As per the suggested alternative method described in https://github.com/AsteroidOS/meta-asteroid/pull/55

In retrospect this is in my opinion the much nicer solution.

### DIFF
--- a/recipes-graphics/ttf-fonts/ttf-asteroid-fonts_1.0.bb
+++ b/recipes-graphics/ttf-fonts/ttf-asteroid-fonts_1.0.bb
@@ -9,8 +9,9 @@ INHIBIT_DEFAULT_DEPS = "1"
 inherit allarch
 inherit fontcache
 
-SRC_URI = "git://github.com/AsteroidOS/asteroid-fonts.git;protocol=https"
-SRCREV = "871b02849988786a76b5948857fdb541417dc94d"
+SRC_URI = "git://github.com/AsteroidOS/asteroid-fonts.git;protocol=https \
+    file://69-emoji.conf"
+SRCREV = "${AUTOREV}"
 S = "${WORKDIR}/git"
 
 FONT_PACKAGES = "ttf-asteroid-fonts"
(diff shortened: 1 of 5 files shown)


FlorentRevest:
While looking into this I noticed that asteroid-launcher is the recipe depending on asteroid fonts. I find it peculiar, it's certainly a bad decision on my side from a long long time ago. I would suggest that we move this dependency to the asteroid-image bbclass

recipes-graphics/ttf-fonts/ttf-asteroid-fonts_1.0.bb
```diff
@@ -9,15 +9,21 @@ INHIBIT_DEFAULT_DEPS = "1"
...
     find ./ -name '*.[to]tf' -exec install -m 0644 {} ${D}/usr/share/fonts \;
+
+    install -d ${D}/home/ceres/.config/fontconfig/conf.d/
+    install -m 644 ${WORKDIR}/69-emoji.conf ${D}/home/ceres/.config/fontconfig/conf.d/
 }
 
-FILES_${PN} += "/usr/share/fonts"
+FILES_${PN} += "/usr/share/fonts /home/ceres/.config/fontconfig/conf.d"
+
+INSANE_SKIP_${PN} += "host-user-contaminated"
```
FlorentRevest, line 29: What was the error that you had to skip here ?

FlorentRevest (approved):
I am also somewhat ambivalent about which one of your two PRs has the best approach. The separated recipe means it will be easier to get automatic updates when they happen in the future, the asteroid-fonts approach means it will be easier to maintain our set of fonts in a working state. Originally I created asteroid-fonts to distribute a bunch of fonts that are not available in OE (for localized characters) but I see that it also contains fonts like Roboto so it already sort of evolved into a mega font repo, I think that's okay. Let's stick with the asteroid-fonts approach then.

recipes-graphics/ttf-fonts/ttf-asteroid-fonts_1.0.bb
MagneFire: The exact warning:
```
WARNING: ttf-asteroid-fonts-1.0-r0 do_package_qa: QA Issue: ttf-asteroid-fonts: /ttf-asteroid-fonts/home/ceres/.config/fontconfig/conf.d/69-emoji.conf is owned by uid 1000, which is the same as the user running bitbake. This may be due to host contamination [host-user-contaminated]
```

I believe this is due to the config file being installed in a `ceres` owned directory. Similarly `lipstick` does the same:
https://github.com/AsteroidOS/meta-asteroid/blob/master/recipes-nemomobile/lipstick/lipstick_git.bb#L42
`lipstick` is the only other recipe to install files in a `ceres` owned directory.

MagneFire: If you prefer, I can add this warning in the recipe as well.

FlorentRevest: I see :)
Just add a comment then and it should be good ;)

MagneFire: Done, thanks for reviewing my PRs :)

PR #57, opened 2021-01-14
Make layer compatible with gatesgarth.

MagneFire:
This PR consists of many recipe changes to fix build issues with the latest gatesgarth.

## Some notes
Some things that I've looked into that we may want to discuss:

*qtbase*: Remove the force GLES2 patch, caused build failure. But seems to work fine on watch. Is this patch really needed? (qtwayland has a similar patch which is still there).

*qofonoext* is actually no longer needed as it is now also in meta-qt5(libqofonoext). Will remove later, needs testing.

*wayland & hybris*: Wayland's patch to remove Wayland-egl was removed in favor of a patch to remove it in libhybris. The main reason behind this is because upstream libhybris does the same.

*tut-dev*: This package seems deprecated? (No updates upstream) Also needs Python 2 (hardcoded to /usr/bin/python2) in order for it to work.

*ofono*: We seem to use a different ofono fork. The fork is 3 years old now, is there a specific reason for using this? I noticed that upstream ofono does not have all the patches from the fork merged, this is the reason for keeping the old one for now.

*mapplauncherd*: There is an important change here that makes all AsteroidOS apps now a shared library. Previously AsteroidOS apps were built as position-independent code. But with glibc >= 2.30 it is no longer possible to use this. The solution/workaround is to build AsteroidOS apps as shared libraries. But this makes it no longer trivial to launcher them as a command(i.e. You can only launch them using the `invoker`).

Of course this is just initial work, future PRs should update other recipes and packages like `mce`. To get rid of `statefs` solving the undefined issue that has plagued this project 🤣

Also, `asteroid-launcher` needs updates to fix some deprecation issues, but those are warnings for now, will come back to that later.

Relevant PRs:
- asteroid: https://github.com/AsteroidOS/asteroid/pull/146
- lipstick: https://github.com/AsteroidOS/lipstick/pull/12
- meta-sturgeon: https://github.com/AsteroidOS/meta-sturgeon-hybris/pull/15

This PR solves https://github.com/AsteroidOS/asteroid/issues/141 and solves https://github.com/AsteroidOS/asteroid/issues/138. Though I should add that if Python 3.9 is installed the build will fail due to xcb, but patches are already available for that too: https://gitlab.freedesktop.org/xorg/proto/xcbproto/-/commit/426ae35bee1fa0fdb8b5120b1dcd20cee6e34512 So once openembedded updates we should be good to go.


recipes-nemomobile/tut-dev/tut-dev_git.bb
```diff
@@ -11,8 +11,10 @@ S = "${WORKDIR}/git"
 
 inherit waf
 
-B = "${S}"
+B = "${WORKDIR}/build"
```
FlorentRevest, line 14: Isn't that the default value of B?

FlorentRevest: https://github.com/openembedded/openembedded-core/blob/master/meta/classes/waf.bbclass#L8

FlorentRevest (approved):
Great! :) I'm glad this removes so many downstream patches.

I know that updating OE is unsexy work but it's really important and appreciated. ;) Thanks MagneFire

FlorentRevest:
> qtbase: Remove the force GLES2 patch, caused build failure. But seems to work fine on watch. Is this patch really needed? (qtwayland has a similar patch which is still there).

Not necessary

> qofonoext is actually no longer needed as it is now also in meta-qt5(libqofonoext). Will remove later, needs testing.

Great :) 

> wayland & hybris: Wayland's patch to remove Wayland-egl was removed in favor of a patch to remove it in libhybris. The main reason behind this is because upstream libhybris does the same.

Sure, I remember starting to upgrade OE a while back and I had to do the same

> tut-dev: This package seems deprecated? (No updates upstream) Also needs Python 2 (hardcoded to /usr/bin/python2) in order for it to work.

Yeah this is unfortunate... But I think we need it as a dependency of some other tools or lib ?

> ofono: We seem to use a different ofono fork. The fork is 3 years old now, is there a specific reason for using this? I noticed that upstream ofono does not have all the patches from the fork merged, this is the reason for keeping the old one for now.

Tbh we can drop it... I was experimenting with having the downstream MTK ofono working on harmony a looong while ago but suddenly stopped when the connectwatch project died and since then I don't think anyone picked this up. So this was left in a very early stage (I think the SIM was detected though) but if this is just maintenance burden now I don't mind dropping it.

> mapplauncherd: There is an important change here that makes all AsteroidOS apps now a shared library. Previously AsteroidOS apps were built as position-independent code. But with glibc >= 2.30 it is no longer possible to use this. The solution/workaround is to build AsteroidOS apps as shared libraries. But this makes it no longer trivial to launcher them as a command(i.e. You can only launch them using the invoker).

Great  :) 

> Of course this is just initial work, future PRs should update other recipes and packages like mce. To get rid of statefs solving the undefined issue that has plagued this project rofl

If this is not a blocker for the gatesgarth to boot I'd say merge this PR as is (you have approval once the minor comment is fixed). We can always update mce separately. And when it comes to statefs... gah :p I was hoping that this would somehow vanish when upgrading OE eheh.

> Also, asteroid-launcher needs updates to fix some deprecation issues, but those are warnings for now, will come back to that later.

Fine

> Relevant PRs:
> 
> asteroid: AsteroidOS/asteroid#146
> lipstick: AsteroidOS/lipstick#12
> meta-sturgeon: AsteroidOS/meta-sturgeon-hybris#15
> This PR solves AsteroidOS/asteroid#141 and solves AsteroidOS/asteroid#138. Though I should add that if Python 3.9 is installed the build will fail due to xcb, but patches are already available for that too: https://gitlab.freedesktop.org/xorg/proto/xcbproto/-/commit/426ae35bee1fa0fdb8b5120b1dcd20cee6e34512 So once openembedded updates we should be good to go.

You have approvals there too so you can merge them after this one. The -hybris layers need a similar change than the one you made for meta-sturgeon and the one jrt made for meta-dory but you have my approval to push directly to the -hybris repos the same set of changes.

recipes-nemomobile/tut-dev/tut-dev_git.bb
MagneFire: This is indeed correct. I have verified this locally. I shall remove this in a subsequent rebase.

MagneFire:
Thanks for reviewing!

I had a look into the dependency train for tut-dev and statefs.

qtaround and cor are the packages that depend on tut-dev.
cor:
 - statefs-qt depends on qtaround and cor.
 - statefs-loader-qt depends on cor.
 - qtaround depends on cor.
qtaround:
- statefs-qt is the only one that depends on qtaround
All these packages are now obsolete.

statefs-qt has the following dependencies:
- lipstick
- buteo-mtp
- timed
- nemo-qml-plugin-contextkit
- asteroid-btsyncd

asteroid-launcher(https://github.com/AsteroidOS/asteroid-launcher/blob/master/qml/quicksettings/QuickSettings.qml#L47) depends on nemo-qml-plugin-contextkit.

Analyzing the code a bit makes it seem as if statefs is only/mainly used for battery status. (don't know why for buteo-mtp and timed just yet.)
It seems that mce-qt is now the de-facto method to get battery information.

So I think that it is possible to remove most of these obsolete packages by updating the packages that depend on statefs-qt. Mainly change the implementation to use mce-qt.

FlorentRevest:
Getting rid of statefs would be awesome but I don't know what buteo-mtp uses it for. I'm assuming this is probably not for the battery status so you would have to dig into that too

PR #58, opened 2021-02-14
Add support for community repositories

MagneFire:
This PR adds the ability for other layers to define a set of packages to install in the package feed for use with `opkg`. An example use is like what I've done in the `meta-games` repo: https://github.com/MagneFire/meta-games/blob/community_repo/conf/layer.conf#L13.

A nice thing is that packages won't be build in the package feed when a layer doesn't exist. So it decouples the community packages from the main repos.

Also, because of this added functionality I have removed the creation of the package feed from the `bitbake asteroid-image` command.
So users building AsteroidOS don't actually build the package feed. This however does mean that an adjustment has to be made on the build server such that after a asteroid-image build the `bitbake asteroid-package-feed` command is used to update the package feed and with it the community packages.

### DIFF
--- a/recipes-asteroid/asteroid-package-feed/asteroid-package-feed.bb
+++ b/recipes-asteroid/asteroid-package-feed/asteroid-package-feed.bb
@@ -1,0 +1,30 @@
+SUMMARY = "Create a package index from the packages included in rootfs and community packages."
+LICENSE = "MIT"
+
+INHIBIT_DEFAULT_DEPS = "1"
+PACKAGES = ""
+
+# A set of packages to build for the package feed but that are not included in rootfs.
+PACKAGE_FEED ?= ""
+
+inherit nopackages
+
+deltask do_fetch
+deltask do_unpack
+deltask do_patch
+deltask do_configure
+deltask do_compile
+deltask do_install
+deltask do_populate_lic
+deltask do_populate_sysroot
+
+do_package_index[nostamp] = "1"
+do_package_index[depends] += "${PACKAGEINDEXDEPS}"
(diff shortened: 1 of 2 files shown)


recipes-asteroid/asteroid-package-feed/asteroid-package-feed.bb
```diff
@@ -0,0 +1,30 @@
+SUMMARY = "Build community packages and create package index from the community packages and rootfs pacakges."
+LICENSE = "MIT"
+
+INHIBIT_DEFAULT_DEPS = "1"
+PACKAGES = ""
+
+# A set of packags to build for the package feed bu not include in rootfs.
```
FlorentRevest, line 7: but that are not included*

recipes-asteroid/asteroid-package-feed/asteroid-package-feed.bb
```diff
@@ -0,0 +1,30 @@
+SUMMARY = "Build community packages and create package index from the community packages and rootfs pacakges."
```
FlorentRevest, line 1: I would reword that as "Create a package index from the packages included in rootfs and community packages"

FlorentRevest (approved):

recipes-asteroid/asteroid-package-feed/asteroid-package-feed.bb
MagneFire: ow, wow. Looks almost as if my keyboard is broken :smile:
Sorry for not checking this...

jrtberlin (approved):

PR #61, opened 2021-04-10
Updates and remove statefs.

MagneFire:
This PR adds the appropriate changes to entirely get rid of `statefs`.

All these PRs and bump SRCREVs are related to battery tracking, which is now part of Mce.
With the recent work towards `gatesgarth` (https://github.com/AsteroidOS/meta-asteroid/pull/57) I noticed that `statefs` was crashing frequently, causing for incorrect battery reporting (on `sturgeon` the battery level was either 99% or 100%).

I don't know if it's wise to merge this all just yet, it works fine on `sturgeon`, but I haven't tested most of these changes on other watches just yet. I'd like to test it on at least a few more watches before the merge, but in my opinion it's at least ready for review/silly mistakes checks :smile:

On an unrelated note: `timed` has a lovely patch that `Fixes build`, this patch should ideally be changed to something better like using the `type_to_cxx` tool to generate the types instead.

This finally fixes https://github.com/AsteroidOS/asteroid/issues/95 and fixes https://github.com/AsteroidOS/asteroid/issues/94.

Other relevant PRs:
- https://github.com/AsteroidOS/mce/pull/9
- https://github.com/AsteroidOS/sensorfw/pull/5
- https://github.com/AsteroidOS/lipstick/pull/13
- https://github.com/AsteroidOS/asteroid-btsyncd/pull/19
- https://github.com/AsteroidOS/asteroid-launcher/pull/48

PR #63, opened 2021-05-09
Port to CMake

MagneFire:
This is part of the excellent work by @PureTryOut https://github.com/AsteroidOS/qml-asteroid/pull/14!

In order to reduce image space used, the built translation files are removed when they were only used for the .desktop file generation.

This PR does not include the needed changes for asteroid-launcher as the relevant PR (https://github.com/AsteroidOS/asteroid-launcher/pull/49) for that one depends on the removal of statefs: https://github.com/AsteroidOS/meta-asteroid/pull/61

I should mention that even though this work now also checks for runtime QML modules. It fails to find all of them. This currently results into a compile time warning, that is only visible when looking at the configure logs.

### DIFF
--- a/recipes-asteroid/asteroid-flashlight/asteroid-flashlight_git.bb
+++ b/recipes-asteroid/asteroid-flashlight/asteroid-flashlight_git.bb
@@ -8,7 +8,11 @@ SRCREV = "${AUTOREV}"
 PR = "r1"
 PV = "+git${SRCPV}"
 S = "${WORKDIR}/git"
-inherit qmake5
+inherit cmake_qt5
 
-DEPENDS += "qml-asteroid qttools-native qtdeclarative-native"
-FILES_${PN} += "/usr/share/translations/"
+DEPENDS += "qml-asteroid asteroid-generate-desktop-native qttools-native qtdeclarative-native"
+
+do_install_append() {
+    # This app only uses translations for the desktop shortcut.
+    rm -rvf ${D}/usr/share/translations/
+}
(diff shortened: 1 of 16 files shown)


recipes-asteroid/asteroid-flashlight/asteroid-flashlight_git.bb
```diff
@@ -8,7 +8,12 @@ SRCREV = "${AUTOREV}"
 PR = "r1"
 PV = "+git${SRCPV}"
 S = "${WORKDIR}/git"
-inherit qmake5
+#inherit qmake5
```
FlorentRevest, line 11: No need to keep this commented

recipes-asteroid/qml-asteroid/qml-asteroid-native_git.bb
```diff
@@ -0,0 +1,17 @@
+SUMMARY = "QML components for AsteroidOS"
```
FlorentRevest, line 1: This recipe would be better named asteroid-generate-desktop-native. The recipe as-is gives the impression that it builds qml-asteroid but it actually does not

recipes-devtools/cmake/extra-cmake-modules_git.bb
```diff
@@ -0,0 +1,15 @@
+UMMARY = "Extra modules and scripts for CMake"
```
FlorentRevest, line 1: issing a letter

FlorentRevest (changes requested):

PR #72, opened 2021-10-29
Add tune-cortexa7.inc for compatability with armv7ve allowing certain…

R0NAM1:
… watches to build


R0NAM1:
File exists in OE-CORE, path was updated and I didn't realize it.

PR #73, opened 2021-10-30
Resolve library conflict between libhybris and wayland

R0NAM1:
According to https://github.com/agherzan/meta-raspberrypi/issues/243#issuecomment-388374552 in Wayland 1.15+ certain libraries that were provided by just libhybris previously are now also provided by Wayland, this causes issues during packaging (Specifically with gstreamer1.0-plugins-base-1.18.5-r0), the above fixes it and allows it to package by only packing the libs from one source.


R0NAM1:
Syntax Strikes again -_-

PR #75, opened 2021-11-07
fix resize artifacts from bootlogo pngs and reduze filesize

eLtMosen:
As discussed on matrix, pngs are now exported to correct size from svg.
This should make them sharper. 
The vertical-center alignment of the logo was off by some pixels and has been corrected.


FlorentRevest:
This is excellent :) Thanks for caring about Asteroid being pixel-perfect! The improvement is very clear. :D

PR #76, opened 2021-11-09
psplash: Add support for animated gifs during boot

MagneFire:
Add support for showing an animated gif located at /usr/share/psplash.gif.
This is enabled by default and can be disabled on a per device basis if this proves problematic on some platforms.

Many credits go to @eLtMosen for providing the animated gifs!
This work is based on https://github.com/tano-systems/psplash-tn/commit/67394ea39573a5f84672b78ffe32ee47f21be410

Demo:

https://user-images.githubusercontent.com/7857908/140948628-c424393a-272f-4ff8-9526-0adcdc000379.mp4

Also worth noting that if a watch needs a different solution the following can be used in a `bbappend` recipe:
```
do_install:append:sturgeon() {
    install -m 0755 ${WORKDIR}/psplash-img-400-220.gif ${D}/usr/share/psplash.gif
}
```
----

Now also distribute the `Disable-double-buffering.patch` patch here instead of the different watch layers (https://github.com/AsteroidOS/meta-ray-hybris/blob/master/recipes-core/psplash/psplash/0001-Disable-double-buffering.patch, https://github.com/AsteroidOS/meta-skipjack-hybris/blob/main/recipes-core/psplash/psplash/0001-Disable-double-buffering.patch, also for `sawfish` (oreo +) and `catfish`).

Additionally, instead of restarting psplash and starting msm-fb-refresher with some delay. We can also start msm-fb-refresher before psplash to solve the garbled display issue (https://github.com/AsteroidOS/asteroid/issues/101#issuecomment-752044885).


FlorentRevest:
Eheh this is very cool :) Let's see if we can come up with more cool animations.

PR #79, opened 2021-11-26
Add Diamonds game to stock images.

MagneFire:
This PR adds the 2048 game to stock AsteroidOS images.
During the past few days many changes were made to the original 2048 game for AsteroidOS.

Thanks to great help from @eLtMosen and others from the matrix chat group we came up with the design shown in the image below.

One remark I have to make is that the name of the app will result it being the first entry in the app launcher. In my opinion this isn't very desirable. So instead I propose to change the name to: `asteroid-diamonds` as the current game rotated the grid by 45 degrees to create a diamond shape. This has the benefit of the game not being the first entry in the list and also it gives it a unique name because the game doesn't look identical to a 2048 clone anymore.

Obviously, before merging we need to also move the repo (https://github.com/MagneFire/asteroid-2048/tree/devel) under the AsteroidOS umbrella. So the code for that one probably also could benefit some reviewing.

Opinions on whether to include this game to stock images are also welcome :smile:

Relevant other PR: https://github.com/AsteroidOS/meta-asteroid-community/pull/7

### DIFF
--- a/classes/asteroid-image.bbclass
+++ b/classes/asteroid-image.bbclass
@@ -7,7 +7,7 @@ IMAGE_FEATURES += "package-management debug-tweaks"
 IMAGE_INSTALL += " \
 kernel-modules base-files base-passwd systemd busybox iproute2 connman pam-plugin-loginuid bluez5 pulseaudio-server openssh-sshd openssh-sftp-server openssh-scp statefs dsme mce ngfd timed sensorfw android-init resize-rootfs mapplauncherd-booster-qtcomponents usb-moded ofono \
 ${@oe.utils.conditional('MACHINE_HAS_WLAN', 'true', 'iproute2 wpa-supplicant connman-client', '', d)} \
-qtgraphicaleffects-qmlplugins supported-languages ttf-asteroid-fonts asteroid-launcher asteroid-calculator asteroid-calendar asteroid-stopwatch asteroid-settings asteroid-timer asteroid-alarmclock asteroid-weather asteroid-music asteroid-btsyncd asteroid-flashlight"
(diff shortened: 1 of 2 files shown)


eLtMosen:
Having played some sessions now, i gotta say i like the fresh variation on the game concept very much.
- Due to the rotation, it is a literal mind twist to adapt from the regular 2048 version which is a nice touch of additional challenge.
- One prominent tactic to solve 2048 is to keep the large numbers in one corner afaiu (at least solved the game that way regulary) and the rotation makes that tactic kind of obvious and invites to follow it by suggesting the lowest square is where things move gravitationally.
- As pointed out by @beroset on matrix, the rotation also helps with preventing false recognition of gestures to e.g. close the game by top down swipe accidentally.
- Not to forget the original idea for rotating the game board, it now resembles the Asteroid Logo very close.

Great work everone!

FlorentRevest:
This is seriously amazing! I love the new design so much, great job guys :D 
The only thing I would maybe change is the remaining colors of the tiles. I see this is still the default color scheme of the original 2048 game and even though it kinda works it could probably be a better fit with the rest of the app if the tones were a bit adapted to this beautiful dark orange background. :) 

Also I'm okay with the Diamond name ;) The repository would need to be created with that name and the recipe adapted.

eLtMosen:
> The only thing I would maybe change is the remaining colors of the tiles. I see this is still the default color scheme of the original 2048 game and even though it kinda works it could probably be a better fit with the rest of the app if the tones were a bit adapted to this beautiful dark orange background. :)

Already on it and did dome attempts, i'll share when i am satisfied with one of them ;)

PR #85, opened 2022-02-02
libhybris: Fix compilation.

MagneFire:
libhybris uses pkgconfig to get the prefix for the wayland-scanner. Unfortunally this causes it to use: /usr/bin/wayland-scanner (host tool!).
We don't want this to happen, instead patch it so that it uses wayland-scanner from the sysroot.

~Additionally move the applying of the other two patches to here too. It should not be needed to apply this separately for every watch. Instead a watch can just clear the patches required for a 'new' libhybris build.~ Bad choice because it's not easy to clear all patches used!

I'll need to undo the above


FlorentRevest:
Is this something that we could also just upstream to the upstream libhybris instead ?

PR #91, opened 2022-03-23
asteroid-launcher: Distribute the asteroid-launcher package as machine specific.

MagneFire:
Every watch defines their own set of input devices for the touchscreen and/or other input.
This makes the package machine specific, we cannot distribute this package identically for all watches (armv7vehf-neon).

This issue has been reported in the past but came back to my mind after @beroset mentioned this issue again (Thanks!).


FlorentRevest:
Ah yeah that's interesting... I think ultimately we should separate the device specific config in a separate package so that both qml-asteroid and asteroid-launcher can be kept generic across all watched (this would make a lot more sense really) but for now that will do! :)

PR #93, opened 2022-04-01
asteroid-machine-config: Regenerate config file when machine variables change.

MagneFire:
Additionally, this removes asteroid-machine-config from the build time dependencies (it's only a runtime dependency). This improves build times as asteroid apps don't need rebuilding every time a machine variable changes (which was the entire reason for this thing's existence :wink: )


FlorentRevest:
Great job :) I think we should also move the device specific asteroid-launcher config into asteroid-machine-config so that it's also not rebuilt for every watch

PR #95, opened 2022-04-09
asteroid-gps: Adds GPS test application.

MagneFire:
It might be handy to actually provide a recipe to build the GPS testing application.


FlorentRevest:
Cool :) Probably we should also build it as part of our community packages ?

PR #98, opened 2022-04-29
Update Bluebinder and add support for the sensorfw binder version.

MagneFire:
This updates the necessary packages so that `sensorfw` can be compiled with binder support.
It splits the `sensorfw` package into two different packages so that a watch can select the default hybris backend (using the `sensors.<codename>.so`) or the binder backend (that utilizes `/dev/hwbinder`).

This requires an update to the `meta-smartwatch` layer since the original `sensorfw` recipe no longer exists.


MagneFire:
I have no idea why, but for some reason everything broke when it comes to this new approach of selecting a `sensorfw` variant.

The general idea was that there doesn't exist a `sensorfw` package (it's virtual) and one selects a package as the `sensorfw` package by means of the `PREFERRED_PROVIDER_sensorfw` variable.

However this does not work. It would always result in "Nothing RPROVIDES 'sensorfw'...". Meaning that the runtime provider wasn't set for some reason.
Now, it turns out that this should never work as it's meant for when there are packages that provide `sensorfw` (so, `sensorfw-hybris` and `sensorfw-binder` would both provide `sensorfw`, which isn't actually the case in this PR).
To solve the latter case, we can utilize the `PROVIDES +=` variable.
This works for building the `sensorfw` recipe. But breaks down when you want to install `sensorfw`. As there isn't a package named `sensorfw`.

I did find another solution to this problem and that is to simply use the `PREFERRED_VERSION_sensorfw` variable and prefix the sensorfw package with a version.
In that case you get the `sensorfw` package but there are multiple versions available.
Selecting the binder variant can be done by means of `PREFERRED_VERSION_sensorfw = "binder%"`.
It seems to work. But I'll need to test it some more.

MagneFire:
Now superseded by https://github.com/AsteroidOS/meta-asteroid/pull/99.

PR #99, opened 2022-05-08
sensorfw: Split sensorfw into a core and hybris.

MagneFire:
This provides a better solution to https://github.com/AsteroidOS/meta-asteroid/pull/98.

Separates the core functionality and HAL specific plugins so that we can have a generic version used by the QEMU version.
The HAL recipe provides additional plugins specifically for the Android HAL backend.

Related: https://github.com/AsteroidOS/meta-smartwatch/pull/81


recipes-nemomobile/sensorfw/sensorfw-hybris-hal-plugins_git.bb
```diff
@@ -0,0 +1,18 @@
+require sensorfw.inc
+
+EXTRA_QMAKEVARS_PRE = "CONFIG+=autohybris"
+
+do_install:append() {
+    # Remove sensorfw core elements.
```
FlorentRevest, line 6: How about
```
rm -rvf ${D}/!(usr) \
        ${D}/usr/!(lib) \
        ${D}/usr/lib/!(sensord-qt5)
```
? Maybe there are even more clever bash tricks but that's where my bash-fu stops :)

recipes-nemomobile/sensorfw/sensorfw-hybris-hal-plugins_git.bb
```diff
@@ -0,0 +1,18 @@
+require sensorfw.inc
+
+EXTRA_QMAKEVARS_PRE = "CONFIG+=autohybris"
```
FlorentRevest, line 3: In that case we should *always* want hybris so maybe we should use CONFIG+=hybris instead ? Iirc this would avoid dynamically checking if hybris is enabled which would silently avoid compiling the hybris plugin if hybris is not detected.

recipes-nemomobile/sensorfw/sensorfw-hybris-hal-plugins_git.bb
```diff
@@ -0,0 +1,18 @@
...
+    rm -rvf ${D}/lib/
+    rm -rvf ${D}/etc/
+    rm -rvf ${D}/usr/share/
+    rm -rvf ${D}/usr/lib/libsensor*
+    rm -rvf ${D}/usr/lib/pkgconfig/
+}
+
+DEPENDS += "sensorfw libhybris"
+
+FILES:${PN} = "/usr/lib /usr/lib/sensord-qt5/*.so /usr/lib/sensord-qt5/testing/*.so"
```
FlorentRevest, line 18: Is that necessary ? I expect all the content of /usr (that is, all the content listed here) would end up in the right recipe ?

recipes-nemomobile/sensorfw/sensorfw-hybris-hal-plugins_git.bb
```diff
@@ -0,0 +1,18 @@
+require sensorfw.inc
+
```
FlorentRevest, line 2: Shouldn't this be marked for machine specific recompilation ?

FlorentRevest, line 2: It would be good to redefine SUMMARY here and clarify what's going on here

recipes-nemomobile/sensorfw/sensorfw_git.bb
```diff
@@ -1,39 +1,9 @@
...
-    sed -i 's:-L/usr/lib ::' ../git/core/hybris.pro
-}
+require sensorfw.inc
 
 do_install:append() {
-    install -d ${D}/etc/sensorfw/ ${D}/lib/systemd/system/multi-user.target.wants/
-    cp ${S}/config/sensord-hybris.conf ${D}/etc/sensorfw/primaryuse.conf
-    cp ../sensorfwd.service ${D}/lib/systemd/system/sensorfwd.service
-    ln -s ../sensorfwd.service ${D}/lib/systemd/system/multi-user.target.wants/
+    rm -rvf ${D}/usr/lib/sensord-qt5/
```
FlorentRevest, line 4: Do we have any other plugin that is not hybris in this directory ?

recipes-nemomobile/sensorfw/sensorfw-hybris-hal-plugins_git.bb
MagneFire: This would be awesome if it worked :wink: 

Using any kind of pattern matching features appears to make it fail to build the package:

```
| DEBUG: Python function extend_recipe_sysroot finished
| DEBUG: Executing shell function do_install
| /home/darrel/AsteroidOS/asteroid-20220512/build/tmp-glibc/work/armv7vehf-neon-oe-linux-gnueabi/sensorfw-hybris-hal-plugins/+gitAUTOINC+2e53901599-r1/temp/run.do_install.3000954: line 166: syntax error near unexpected token `('
| WARNING: /home/darrel/AsteroidOS/asteroid-20220512/build/tmp-glibc/work/armv7vehf-neon-oe-linux-gnueabi/sensorfw-hybris-hal-plugins/+gitAUTOINC+2e53901599-r1/temp/run.do_install.3000954:152 exit 2 from 'export systemd_user_unitdir="/usr/lib/systemd/user"'
```

From what I can tell, pattern matching this way is a feature exclusive to bash. I could not find which command interpreter is used by bitbake.

A potential solution I found is shown here: https://unix.stackexchange.com/questions/670719/pattern-matching-in-if-block-in-alpines-sh
But in my opinion this will never look as neat as the one you showed.

So it looks like the current approach of explicitly deleting the core files might be the best way? Or do you know if I'm missing something here?

recipes-nemomobile/sensorfw/sensorfw-hybris-hal-plugins_git.bb
MagneFire: Good idea, but won't work for now since the `hybris` target fails to compile currently.
Is it ok for you if I fix this in a next PR (https://github.com/AsteroidOS/meta-asteroid/pull/98 but using the current work as the base)?

recipes-nemomobile/sensorfw/sensorfw-hybris-hal-plugins_git.bb
MagneFire: Good point! `/usr/lib` isn't required here.

recipes-nemomobile/sensorfw/sensorfw-hybris-hal-plugins_git.bb
MagneFire: We can. But I'm not entirely sure if we need it.
My thought process is as follows: All watches that don't need a patched `sensorfw` could use the generic `armv7vehf-neon` while the watches that need a patched version can use the one specific to a machine.
This would mean that https://github.com/AsteroidOS/meta-smartwatch/pull/81 needs the `PACKAGE_ARCH = "${MACHINE_ARCH}"` applied for some watches.

Or would it be better to make every version specific to a watch? Just apply `PACKAGE_ARCH = "${MACHINE_ARCH}"` here. Like your comment.

recipes-nemomobile/sensorfw/sensorfw_git.bb
MagneFire: Yes! Good point!

recipes-nemomobile/sensorfw/sensorfw-hybris-hal-plugins_git.bb
MagneFire: Found a solution using `find`.

recipes-nemomobile/sensorfw/sensorfw-hybris-hal-plugins_git.bb
MagneFire: Now integrated in this PR.

recipes-nemomobile/sensorfw/sensorfw-hybris-hal-plugins_git.bb
MagneFire: Added anyway since it solves another issue where `libhybrissensorfw*` is installed in the deploy directory by the hal and the binder version. It resulted in the following issue:
```
ERROR: sensorfw-hybris-hal-plugins-+gitAUTOINC+2e53901599-r1 do_package_write_ipk: The recipe sensorfw-hybris-hal-plugins is trying to install files into a shared area when those files already exist. Those files and their manifest location are:
  /home/darrel/AsteroidOS/asteroid-20220512/build/tmp-glibc/deploy/ipk/sturgeon/libhybrissensorfw-qt5-dev_+git0+2e53901599-r1_sturgeon.ipk
    (matched in manifest-sturgeon-sensorfw-hybris-binder-plugins.package_write_ipk)
  /home/darrel/AsteroidOS/asteroid-20220512/build/tmp-glibc/deploy/ipk/sturgeon/libhybrissensorfw-qt5-1_+git0+2e53901599-r1_sturgeon.ipk
    (matched in manifest-sturgeon-sensorfw-hybris-binder-plugins.package_write_ipk)
  /home/darrel/AsteroidOS/asteroid-20220512/build/tmp-glibc/deploy/ipk/sturgeon/libhybrissensorfw-qt5-dbg_+git0+2e53901599-r1_sturgeon.ipk
    (matched in manifest-sturgeon-sensorfw-hybris-binder-plugins.package_write_ipk)
  /home/darrel/AsteroidOS/asteroid-20220512/build/tmp-glibc/deploy/ipk/sturgeon/libhybrissensorfw-qt5-src_+git0+2e53901599-r1_sturgeon.ipk
    (matched in manifest-sturgeon-sensorfw-hybris-binder-plugins.package_write_ipk)
Please verify which recipe should provide the above files.
```

There are two solutions to this:
- Use `SSTATE_DUPWHITELIST = "${DEPLOY_DIR}/"` to allow the dupplicate files to exist. This probably introduces an issue with the package manager though since we can't reliably say what the ipk targets (hal or binder?).
- Make the package machine specific. Only one of the variants should ever be installed for a specific target. This also removes the need for making the package machine specific when a platform overrides it.

FlorentRevest:
Uh sorry that the OE update now created conflicts

recipes-nemomobile/sensorfw/sensorfw-hybris-hal-plugins_git.bb
FlorentRevest: What about a third solution where you'd just rename the libhybrissensorfw* files to libhybrisbindersensorfw* in the binder recipe ? Then both packages should be able to be installed in parallel, no ? Maybe the sensorfw config would need a slight variation to load the binder file but this should be manageable

MagneFire: So, it seems that your proposed solution will not work, since the simply renaming the .so file to something else. This results in the exact same error (100% identical, the .ipk has the exact same name still).

The reason this does not work, is because the package is renamed by Yocto from `sensorfw-hybris-hal-plugins` (and `sensorfw-hybris-binder-plugins`) to `libhybrissensorfw-qt5-1`.
The reason this happens has something to do with how Yocto generates package names. Depending on the contents of the package (https://github.com/openembedded/openembedded-core/blob/master/meta/classes/debian.bbclass#L112) it will use `objdump` if it's a library to get the `SONAME` from the .so. Which is as follows:
```
arm-linux-gnueabi-objdump -p ..../build/tmp-glibc/work/beluga-oe-linux-gnueabi/sensorfw-hybris-binder-plugins/+gitAUTOINC+2e53901599-r1/packages-split/sensorfw-hybris-binder-plugins/usr/lib/libhybrisbindersensorfw-qt5.so.1.0.0

....

Dynamic Section:
  NEEDED               libsensordatatypes-qt5.so.1
  NEEDED               libgbinder.so.1
  NEEDED               libglib-2.0.so.0
  NEEDED               libQt5Core.so.5
  NEEDED               libstdc++.so.6
  NEEDED               libgcc_s.so.1
  NEEDED               libc.so.6
  SONAME               libhybrissensorfw-qt5.so.1
  INIT                 0x00003ca0
  FINI                 0x0000d630
  INIT_ARRAY           0x0001fdcc
  INIT_ARRAYSZ         0x00000004
  FINI_ARRAY           0x0001fdd0
  FINI_ARRAYSZ         0x00000004
  GNU_HASH             0x00000138
....
```

Then it simply renames the package in the `log.do_package` stage: `NOTE: debian: renaming sensorfw-hybris-binder-plugins to libhybrissensorfw-qt5-1`.
Which also results in a package renaming in the `log.do_package_write_ipk`: `NOTE: package name mapping done: sensorfw-hybris-binder-plugins -> libhybrissensorfw-qt5-1`. (https://github.com/openembedded/openembedded-core/blob/master/meta/classes/package.bbclass#L610)

MagneFire: Now solved by setting `DEBIANNAME` so that the package isn't renamed.

PR #102, opened 2022-06-04
ngfd-plugin-droid-vibrator: Fix dependency issue.

MagneFire:
The ngfd droid backend uses the Android headers (version specifically) to figure out which Android API to use.
These headers existed as part of the libhybris dependency, but a change of machine would not always trigger a sysroot update because there's no hard dependency on the android-headers package.

This fixes the issue where sometimes the vibrator would not work for some watches.

This issue was reported by @dodoradio, https://github.com/AsteroidOS/meta-smartwatch/issues/90 and https://github.com/AsteroidOS/meta-smartwatch/issues/91.


FlorentRevest:
Ugh another one of these :/ great find though!

PR #104, opened 2022-07-26
Add more information to os-release

beroset:
This adds several useful fields to /etc/os-release so that, for
instance, on sturgeon, the new values look like this:
BUILD_ID="20220726172548"
DEFAULT_HOSTNAME="sturgeon"
LOGO=logo-asteroidos
HOME_URL="https://asteroidos.org/"

These fields, their encodings and meanings are documented here:
https://www.freedesktop.org/software/systemd/man/os-release.html

### DIFF
--- a/recipes-core/os-release/os-release.bbappend
+++ b/recipes-core/os-release/os-release.bbappend
@@ -1,0 +1,6 @@
+OS_RELEASE_FIELDS += "\
+    BUILD_ID LOGO HOME_URL \
+"
+OS_RELEASE_UNQUOTED_FIELDS += " LOGO"
+HOME_URL = "https://asteroidos.org/"
+LOGO = "logo-asteroidos"
--- a/recipes-asteroid/asteroid-machine-config/asteroid-machine-config_1.0.bb
+++ b/recipes-asteroid/asteroid-machine-config/asteroid-machine-config_1.0.bb
@@ -10,10 +10,14 @@ def display_variables(d):
 def capability_variables(d):
     return ["MACHINE_HAS_WLAN", "MACHINE_HAS_SPEAKER"]
 
+def identity_variables(d):
+    return ["MACHINE"]
+
 python do_generate_config () {
     machine_vars = {
         "Display": display_variables(d),
-        "Capabilities": capability_variables(d)
+        "Capabilities": capability_variables(d),
+        "Identity": identity_variables(d)
     }
 
     from configparser import ConfigParser
@@ -26,6 +30,8 @@ python do_generate_config () {
             v = var[len("MACHINE_"):]
             if v.startswith(key.upper() + "_"):
                 v = v[len(key.upper() + "_"):]
+            if not v:
+                v = var
             if d.getVar(var) is not None:
                 config.set(key, v, str(d.getVar(var)))
 


beroset:
Note that even though there is now a machine-specific entry (DEFAULT_HOSTNAME), this package is build under `all` but that is a problem I wasn't sure how to solve.

beroset:
This is a preliminary step to help address this issue:  https://github.com/AsteroidOS/meta-asteroid/pull/104

FlorentRevest:
> Note that even though there is now a machine-specific entry (DEFAULT_HOSTNAME), this package is build under all but that is a problem I wasn't sure how to solve.

Doesn't `PACKAGE_ARCH = "${MACHINE_ARCH}"` solve this ?

Alternatively, I'm not sure how useful `DEFAULT_HOSTNAME` really is and even less sure how useful it is to have it set as the actual hostname the image ships in by default. I guess my point is, it wouldn't be too bad to ship 
`DEFAULT_HOSTNAME="asteroid"` because it's meant as a fallback anyway.

Either way is fine by me really. It feels a little bit wrong conceptually speaking to have an os-release that depends on the MACHINE but I guess this wouldn't be the biggest hack in asteroid anyway... :)

beroset:
Is there some other way that an app running on the watch get the machine arch?  If so, we could remove the `DEFAULT_HOSTNAME` entirely and use that other mechanism.  I searched but did not find another way.

FlorentRevest:
It feels to me that relying on the `DEFAULT_HOSTNAME` variable to define a codename is a hack because there is no fundamental reason why the hostname or the default hostname should be the watch codename. (you refer to the os-release format, it's pretty clear from it that this field isn't meant to accommodate this :) )

I think that watch codenames are higher level concepts that really only belong in Asteroid and make more sense as part of the watch configuration files we already ship and read here https://github.com/AsteroidOS/qml-asteroid/blob/master/src/utils/src/deviceinfo.cpp#L21

We already generate this file dynamically in a machine dependent way, it seems like a pretty natural to stuff a watch codename definition based on MACHINE https://github.com/AsteroidOS/meta-asteroid/blob/08e2743069f9e5cd7699e8627564bc0556a1bcdf/recipes-asteroid/asteroid-machine-config/asteroid-machine-config_1.0.bb#L13 and export it through qml-asteroid.

This way we keep all the machine specific configs in one machine specific package and file and don't need to hack our os-release, what do you think ?

beroset:
This implements a better solution with the machine name created and populated during build and placed in /etc/machine.conf instead of using the DEFAULT_HOSTNAME hack.  Now all os-release files will get only information that does _not_ depend on the particular watch model.

recipes-core/os-release/os-release.bbappend
```diff
@@ -0,0 +1,7 @@
+PACKAGE_ARCH = "${MACHINE_ARCH}"
```
FlorentRevest, line 1: Now we no longer need this line since the os-release package isn't machine dependent :)

FlorentRevest (changes requested):

recipes-core/os-release/os-release.bbappend
beroset: Right!  I removed that and also DEFAULT_HOSTNAME from the list of tags.

PR #105, opened 2022-07-27
Update to newer version of extra-cmake-modules

beroset:
This is done as a first step in updating nemo-qml-plugin-calendar, and
is also useful for updating other recipes in the future.


beroset:
Note that this will also require changes to the `CMakeLists.txt` of the following projects to replace the old `ECMQMLModules` with the newer `ECMFindQmlModule`.  That change will need to be made in the following projects:

- [ ] asteroid-alarmclock
- [ ] asteroid-btsyncd
- [ ] asteroid-calendar
- [ ] asteroid-compass
- [ ] asteroid-diamonds
- [ ] asteroid-flashlight
- [ ] asteroid-hrm
- [ ] asteroid-music
- [ ] asteroid-settings
- [ ] asteroid-stopwatch
- [ ] asteroid-timer
- [ ] asteroid-weather
- [ ] qml-asteroid

PR #106, opened 2022-07-27
Update libaccounts-qt to version 1.16

beroset:
This is one of the steps required for an update to a working calendar
import for the watch.


beroset:
The next step for fixing `.ics` file imports on the watch is to update `mkcal` which requires and updated version of `libaccounts-qt5` which this PR provides.

PR #109, opened 2022-09-04
Add the usb-moded file location for serial number

beroset:
This patches nemo-qml-plugin-systemsettings to allow that plugin to
correctly fetch the serial number for AsteroidOS and parallels a change
already made to upstream in commit
96b6ae08c6ce47dbcb6625227905eda21383e2bc.

### DIFF
--- a/recipes-nemomobile/nemo-qml-plugins/nemo-qml-plugin-systemsettings/0006-Add-file-to-the-places-searched-for-a-serial-number.patch
+++ b/recipes-nemomobile/nemo-qml-plugins/nemo-qml-plugin-systemsettings/0006-Add-file-to-the-places-searched-for-a-serial-number.patch
@@ -1,0 +1,25 @@
+From 1c61ec5a3457047987ca5ae709762482269d0ea4 Mon Sep 17 00:00:00 2001
+From: =?UTF-8?q?Tomi=20Lepp=C3=A4nen?= <tomi.leppanen@jolla.com>
+Date: Fri, 16 Nov 2018 10:41:00 +0200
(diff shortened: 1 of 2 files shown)


beroset:
This is a complementary change to https://github.com/AsteroidOS/asteroid-settings/pull/52 and helps address https://github.com/AsteroidOS/asteroid-settings/issues/33 by allowing the About page to fetch the serial number.

FlorentRevest:
Have you considered updating nemo-qml-plugin-systemsettings instead ?

Also, to keep the attribution correct, could you just apply this patch: 
https://github.com/sailfishos/nemo-qml-plugin-systemsettings/commit/96b6ae08c6ce47dbcb6625227905eda21383e2bc.patch ? (a git am under devtool should do iiuc)

beroset:
I prepared the patch before I discovered the upstream patch, but your idea is good.  I'll redo this.

beroset:
Actually, I can't do that.  The upstream patch cannot apply cleanly because the context lines are different.  I think we may be stuck with this until we update our version of nemo-qml-plugin-systemsettings to something more recent.

FlorentRevest:
It's quite usual that a patch doesn't apply cleanly but if under devtool you use git am and then fix the conflict manually (most likely in this case a `patch -p1 < 96b6ae08c6ce47dbcb6625227905eda21383e2bc.patch` and `git am --continue` you'll get a commit with the right attribution.

Alternatively, you could edit the patch file to set the right "From:"

beroset:
OK, I think this will do it.

FlorentRevest:
Thank you!

PR #110, opened 2022-09-05
asteroid-compass: Keep translation files.

MagneFire:
Since the recent redesign we now add a calibration text which can be translated.


FlorentRevest:
Good catch :)

PR #114, opened 2022-09-12
qt5-qpa-hwcomposer-plugin: Add Always-on-Display support for hwc2.

MagneFire:
This adds support for Always-on-Display on hwc2 devices (Android 9+).

Additionally it adds support for the use of the PowerHAL on older hwc versions.

Signed-off-by: Darrel Griët <dgriet@gmail.com>


FlorentRevest:
Very cool! :)

PR #119, opened 2022-11-12
Create default virtual keyboard for watch

beroset:
This creates a virtual keyboard that has larger dimensions and is easier to see and use on a watch.


beroset:
This is a modification that allow the virtual keyboard to be actually usable on the watch.  

keyboard

FlorentRevest:
Patching the default style and layout is an option but it makes upgrading qtvirtualkeyboard and changing the style harder to do in the future.

Have you considered using `QT_VIRTUALKEYBOARD_STYLE` and `QT_VIRTUALKEYBOARD_LAYOUT_PATH` https://doc.qt.io/qt-6/technical-guide.html#adding-custom-styles ?

If we could extract our custom style and layout into a new repository and distribute it as a separate package, it would be a lot easier to maintain. Especially as I anticipate more style changes following this one. :)

If you don't feel like doing it, let me know and I can also merge this PR since it's technically still a step in the right direction. (and since the existing patch I wrote already did what you do now... :) )

beroset:
I will try to extract this into a separate package, and maybe clean it up a bit.

beroset:
I've created a separate package.  https://github.com/beroset/asteroid-virtualkeyboard

How shall we handle this?

PR #120, opened 2022-11-22
Add packages for NFC support.

MagneFire:
This PR adds the packages to support NFC. It has been testen on `catfish` and `beluga`.
It does not work on `catfish` because of https://github.com/mer-hybris/libncicore/issues/21 (A ST21NFCD which is a  NCI 2.0 chip).
It appears to work fine on `beluga`.

The packages are not installed by default as no app within AsteroidOS would have a direct benefit from it, we could reconsider this in the future of course.


FlorentRevest:
This is *very* exciting. Thanks a lot once again MagneFire, super cool work

FlorentRevest:
As for whether this should be included in the default image. I don't think it's that heavy so why not.

But also we could use device features to figure out what to install, if the device supports nfc then we would install the stack. This would make it easier for a watch to opt into NFC support rather than installing all sorts of dependencies (we'd have to keep the lists in sync across watch layers which is not the end of the world but also not ideal)

Anyway, as you said, this can be reconsidered later.

PR #121, opened 2022-11-25
Add swclock-offset recipe.

MagneFire:
This PR adds the recipe for swclock-offset (https://gitlab.com/postmarketOS/swclock-offset).
It's purpose is to restore the system time from a non-writeable RTC by using a cached file. Which appears to be an issue on some Qualcomm based watches.

I've also made a PR upstream, but we can use this until that one is merged:
https://gitlab.com/postmarketOS/swclock-offset/-/merge_requests/3

Thanks for @dodoradio for his initial work on this!

### DIFF
--- a/recipes-android/swclock-offset/swclock-offset_git.bb
+++ b/recipes-android/swclock-offset/swclock-offset_git.bb
@@ -1,0 +1,26 @@
+SUMMARY = "Synchronize time from a non-writable real-time clock (RTC) using a cached file."
+HOMEPAGE = "https://gitlab.com/postmarketOS/swclock-offset"
+LICENSE = "GPL-3.0-only"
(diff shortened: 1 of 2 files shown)


recipes-android/swclock-offset/swclock-offset_git.bb
```diff
@@ -0,0 +1,26 @@
...
+SYSTEMD_PACKAGES = "${PN}"
+SYSTEMD_SERVICE:${PN} = "swclock-offset.service"
+
+do_configure[noexec] = "1"
+do_compile[noexec] = "1"
+
+do_install() {
+	oe_runmake install DESTDIR=${D} PREFIX=${prefix}
+
+    install -d ${D}${systemd_unitdir}/system
```
FlorentRevest, line 22: Indentation is not uniform here, otherwise looks good to me feel free to merge after a fix

FlorentRevest (approved):

PR #123, opened 2022-12-14
geoclue-provider-hybris: hal: Handle not being able to set the SUPL server.

MagneFire:
https://github.com/mer-hybris/geoclue-providers-hybris/commit/e982cf94047993e5868d44307f4a38f1b89500ea added support for overriding/setting the SUPL server. This could however lead to a crash when AGPS isn't actually available. Handle this by returning an error instead of crashing.

Signed-off-by: Darrel Griët <dgriet@gmail.com>

This patch will be dropped when the upstream PR is merged: https://github.com/mer-hybris/geoclue-providers-hybris/pull/44


FlorentRevest:
Excellent. And thanks for sending a PR upstream too, that's how we stay good citizens, way too go.

PR #125, opened 2022-12-26
Enable proper qmake, Qt Linguist, moc/uic/rcc, and CMake support in SDKs

dv1:
This allows for building asteroid-btsyncd, asteroid-launcher etc. with SDKs generated by Yocto via calls like:

    bitbake -c populate_sdk asteroid-image

### DIFF
--- a/classes/asteroid-image.bbclass
+++ b/classes/asteroid-image.bbclass
@@ -2,6 +2,9 @@ inherit core-image
 inherit extrausers
 LICENSE = "GPL-2.0-only"
 
+# Inherit this to be able to produce OE SDKs that are fully capable of building Qt5 code
+inherit populate_sdk_qt5
+
 IMAGE_FEATURES += "package-management debug-tweaks"
 
 IMAGE_INSTALL += " \
--- a/recipes-qt/packagegroups/nativesdk-packagegroup-qt5-toolchain-host.bbappend
+++ b/recipes-qt/packagegroups/nativesdk-packagegroup-qt5-toolchain-host.bbappend
@@ -1,3 +1,13 @@
+# - Add nativesdk-cmake to enable CMake support in SDKs.
+#   Newer Qt projects use CMake instead of qmake.
+# - Add nativesdk-qtdeclarative-tools for QML content.
+# - Add nativesdk-qttools-tools to allow for including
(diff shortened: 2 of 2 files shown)


conf/distro/asteroid.conf
```diff
@@ -41,3 +41,12 @@ PACKAGE_FEED_BASE_PATHS = "ipk"
...
 SKIP_META_GNOME_SANITY_CHECK = "1"
+
+# - Add nativesdk-qttools-tools to allow for building
+#   Qt5 code with the SDK. Note that the SDK must be build
+#   based on an image that inherits "populate_sdk_qt5".
+# - Add nativesdk-qttools-dev as well, otherwise the CMake
+#   Qt5LinguistTools scripts are not included in an SDK.
+# - Add nativesdk-cmake to enable CMake support in SDKs.
+#   Newer Qt projects use CMake instead of qmake.
+TOOLCHAIN_HOST_TASK:append = " nativesdk-cmake nativesdk-qttools-tools nativesdk-qttools-dev "
```
dv1, line 52: This is one part where I am uncertain. I do think that this belongs in the distro config, but I do not know if this should use `TOOLCHAIN_HOST_TASK ?=` instead of `TOOLCHAIN_HOST_TASK:append =`.

otavio: This append is correct but belongs to the image where you added the inherit, as we shouldn't force it in all images.

dv1: @otavio There is `asteroid-image-dev`. You would add both the `inherit` and the `TOOLCHAIN_HOST_TASK:append` lines to that image?

otavio: Yes as this is related to the image itself. Another option is a bbappend for the Qt SDK packagegroup.

dv1:
@otavio Fixed. Moved the extra nativesdk packages to the `nativesdk-packagegroup-qt5-toolchain-host` bbappend.

otavio (approved):

FlorentRevest:
This is very nice thanks a lot for this addition! We should use this to generate SDKs too now. Would you mind updating the "Creating an asteroid app" documentation page as well ?

PR #126, opened 2023-01-06
Add cli11 to SDK for building lcd_tools

beroset:
The previous version of SDK did not contain the CLI11 header-only library necessary to build lcd-tools


MagneFire:
Looks good to me, thanks!

FlorentRevest:
We should actually stop maintaining this manual list of dependencies and just use https://github.com/AsteroidOS/meta-asteroid/pull/125 which iiuc includes anything necessary to build anything in asteroid-image

PR #127, opened 2023-01-21
Update package archs

Doomsdayrs:
Strips PACKAGE_FEED_ARCHS to their respective architectures.

Due to a recent update, device specific packages no longer exist.  Throwing a list of errors with opkg.
This fixes the issue by stripping the arch list, and then adding in armv7at2hf-neon.


FlorentRevest:
Thanks for the PR :) 

Unfortunately, this does not seem to be the correct fix. We should rather restore the missing device-specific package feeds. Some packages are device specific and they should be distributed that way.

@beroset seems to be on to something, he found that only the last built watch (wren) has a package feed https://github.com/AsteroidOS/asteroid/issues/235 so this must have something to do with our nightlies build script.

@jrtberlin or @MagneFire care to look into it ?

PR #129, opened 2023-01-25
Use updated upstream and remove patches

beroset:
The changes from the patchfiles were successfully merged into upstream, so they are no longer needed.  This updates the recipe accordingly.


FlorentRevest:
Fantastic :) thank you very much!

PR #130, opened 2023-01-30
Changes to asteroid-image-dev

dv1:
These commits remove dbg-pkgs dev-pkgs to reduce image size, and add htop, bash, and extra Bluetooth tools, which are useful for development and debugging. To quote the first commit's message:

dbg-pkgs and dev-pkgs add ALL dbg and dev packages to the image. The
result is too large for some watches. And, in particular, adding dev
packages to the rootfs of a smartwatch is of very questionable use,
since this is usually done when development tools such as make, cmake,
gcc etc. are to be run on the target itself. This is unlikely to be
done on a smartwatch.

### DIFF
--- a/recipes-core/image/asteroid-image-dev.bb
+++ b/recipes-core/image/asteroid-image-dev.bb
@@ -1,3 +1,3 @@
 inherit asteroid-image
 DESCRIPTION = "Asteroid image for developers, contains development and debugging tools"
-EXTRA_IMAGE_FEATURES += "dbg-pkgs dev-pkgs tools-sdk tools-debug tools-testapps debug-tweaks"
+EXTRA_IMAGE_FEATURES += "tools-debug tools-testapps debug-tweaks"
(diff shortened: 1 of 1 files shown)


otavio (approved):

beroset:
Might it make sense to make these changes for all targets *except* for the qemux86 target?

dodoradio:
I agree with beroset here - I think that it would be useful to be able to run builds directly inside of qemu virtual machine

dodoradio:
we already have some inconsistencies between images, such as the lack of `debug-ramdisk` built into the boot image on smelt. it makes sense to tune down the programs installed on platforms where storage is a limitation, and keep them on platforms like qemu where it isn't. In any case, the qemu image is only used by very hardcore developers and any confusion should be easily resolved with a tiny bit of documentation

dv1:
@dodoradio How about creating a separate `asteroid-image-sdk` then? @otavio , this would be the correct image for the entire compiler and such, right?

FlorentRevest:
Sorry for the late comment here.

I agree that it doesn't make sense for an image to mean different things depending on the MACHINE. I consider the case of smelt as an unfortunate but required hack to accommodate the platform requirements.

Like @dv1, I also think that adding a new image recipe is the better way to go forward. Then, there's no surprise for the user. Adding an image target is cheap and means we don't have to make monobuttockular compromises, we can have each image really specialize in one specific aspect.

PR #131, opened 2023-02-05
Add bluetooth-rfkill-event recipe.

### LINKED ISSUE #114: Same Bluetooth MAC used for multiple watches
At least Lenok and Bass show the same BT-MAC when observed.
Its `43:43:A0:12:1F:AC`.
This makes it hard to pair those watches in succession.

@MagneFire noted on matrix that in some ports BT MAC
> are generally restored from the persist or firnware partition.
> Basically all watches that have support for Bluetooth and don't use the bluebinder might have this issue. sturgeon is amongst them

MagneFire:
bluetooth-rfkill-event can be used to call brcm-patchram-plus whenever the Bluetooth device becomes available via rfkill.
It includes the ability to set the Bluetooth address as well.
Some minor additions have been made to store the randomized Bluetooth address in case there isn't one stored.
It also needed some patching so that we can supply our own rfkill device.

Eventually I'd like to remove the special `brcm-patchram-plus` compile time macros and allow https://github.com/AsteroidOS/brcm-patchram-plus/blob/master/src/main.c#L263 to be configurable via the config files.

See https://github.com/AsteroidOS/meta-smartwatch/pull/155 for device specific changes that need to be applied.

This fixes https://github.com/AsteroidOS/meta-smartwatch/issues/114.

### DIFF
--- a/recipes-core/bluetooth-rfkill-event/bluetooth-rfkill-event-configs_git.bb
+++ b/recipes-core/bluetooth-rfkill-event/bluetooth-rfkill-event-configs_git.bb
@@ -1,0 +1,19 @@
+DESCRIPTION = "Provides the device specific configuration files used by bluetooth-rfkill-daemon"
+PR = "r0"
+LICENSE = "Apache-2.0"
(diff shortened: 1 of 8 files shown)


FlorentRevest:
Woah, that's very nice. Good finding here :) And I like the patches too, good stuff.

PR #137, opened 2023-03-05
initramfs-scripts-android: Wait on the sdcard partition rather than mmcblk0

FlorentRevest:
The sdcard partition can be on another block device and there's no point hard-coding a device when we have $sdcard_partition at our disposal anyway.


MagneFire:
Sorry for not noticing this PR before...

This looks like a sane change, thank you!

PR #138, opened 2023-03-23
Update libconnman-qt to version 1.3.2 and fix bug

beroset:
The asteroid-weather-glow watchface was not working any longer (see https://github.com/AsteroidOS/asteroid/issues/243 ), so this fixes that problem by updating the underlying libconnman-qt to version 1.3.2 from 1.2.46 and adds a necessary patch to wait for the dbus call  to finish before returning from the constructor for NetworkTechnology  objects.


FlorentRevest:
I merged this a bit quickly but had some second thoughts afterward ahah. It looks like you imported a few commits from the libconnman-qt repository as patches into our layer but have you considered bumping the recipe's `SRCREV` instead ? Then we could just fetch these commits from the upstream repo and only ship the additional patches which we actually wrote and maintain

beroset:
I did consider moving the `SRCREV` instead, but was unsuccessful when I tried that approach.  I'll try again.

PR #140, opened 2023-03-26
Amend fix for libconnman-qt5 to bump version

beroset:
The previous fix pulled in the patches from the newer version and applied them, but this does the same thing in a neater way by changing SRCREV and only applying a single patch.


FlorentRevest:
Beautiful!
Thank you 😄

PR #141, opened 2023-03-29
Register dbus types in NetworkTechnology

beroset:
The libconnman-qt5 code's NetworkTechnology object, used by the analog-weather-glow watchface, sometimes attempted to use dbus types that were not yet registered.  This fixes that and is the final piece to fix https://github.com/AsteroidOS/asteroid/issues/243 by registering the types in every version of the NetworkTechnology constructor rather than only within the NetworkManager constructor.


MagneFire:
Looks good, thank you!

Would it make sense to attempt to upstream these changes too to https://github.com/sailfishos/libconnman-qt?

beroset:
Yes, upstreaming these changes is definitely on my "to-do" list.

PR #142, opened 2023-03-29
lcd-tools: remove `COMPATIBLE_MACHINE`

dodoradio:
koi has now been added as to meta-smartwatch. support for it has existed in lcd-tools for a while now, so now seems like an appropriate time to declare it compatible.


MagneFire:
Hmm, PR looks fine but the `COMPATIBLE_MACHINE` part might not be needed since we already perform a runtime check for compatibility.

What's your opinion on this?

dodoradio:
yep, seems reasonable. The tool won't be installed on unsupported platforms by default, and it won't break anything if it is installed. Sounds good to me - should I remake this PR to remove the `COMPATIBLE_MACHINE` line?

MagneFire:
Looks good, thanks for the quick update!

PR #145, opened 2023-04-05
qttools: Add support for mapping the MenuKB key.

MagneFire:
This adds support for mapping the MenuKB key (keycode 139).

It essentially fixes the second button on `catfish`


beroset (approved):
Tested on `catfish` and verified working.

FlorentRevest:
Sweet! :) Is this something that could be contributed upstream ?

PR #147, opened 2023-04-07
Add polkit to allow time change

beroset:
This PR is a prerequisite to being able to apply https://github.com/AsteroidOS/asteroid-settings/pull/63 by adding `polkit` and the polkit rule that allows the `ceres` user to change the time, date, and timezone.


beroset:
**Note to reviewers**: when `DISTRO_FEATURES` is changed, you must wipe the entire `tmp-glibc` directory to rebuild.  This was suggested to me in an error message when I was trying to get it to compile without doing that first and proved to be a critical step in actually getting a clean build.

FlorentRevest:
Ugh, I'm not so happy about adding such a big system daemon to our base image buuuut........ if it's a step in the direction of getting rid of timed.... well... ok uhuh

PR #148, opened 2023-04-09
ttf-asteroid-fonts: Fix license checksum.

MagneFire:
Stock images are now shipped with more fonts (https://github.com/AsteroidOS/asteroid-fonts/commit/341a7ac195ed953c0843b72cc98db9d539a5c993), this resulted into an update of the license text used in the readme.

Currently, `Arkhip` uses a license that's not available in the SPDX License list. We might have to alter this at some point as well.


MagneFire:
`Arkhip` has been removed in `asteroid-fonts` resolving the license issue.

PR #151, opened 2023-04-20
android-init: Fix install path of the android-init.service.

MagneFire:
https://github.com/AsteroidOS/meta-asteroid/commit/06dc67571b21cb2f6ec3e2924cd06a9b354e073d introduced a change where the android-init service file would install to a location using the Yocto system package variables. This however turned out to be the incorrect variable.

The `systemd_unitdir` variable is not the correct path to install systemd unit files (it represents the /lib/systemd/ path).
Instead use `systemd_system_unitdir` variable to install the systemd unit file into the correct path (/lib/systemd/system/).


beroset (commented):
This fixed a problem for me in which `catfish` would boot to the logo screen but no further.

PR #152, opened 2023-04-20
ttf-asteroid-fonts: Update license checksum.

MagneFire:
Stock images are now shipped with more fonts as part of a new stock watchface (https://github.com/AsteroidOS/asteroid-fonts/pull/16), this resulted into an update of the license text used in the readme.


FlorentRevest:
When you make changes like this one which are "obviously non-controversial", and by that I mean, it's 99% unlikely that a reviewer would say "no, let's do this otherwise", please feel free to merge them straight away. This keeps the review load a bit lower and keeps the project velocity a bit higher ;) You have the rights to merge your own PRs, you can apply it with common-sense ;) And for the 1 remaining percent, cleaning up the mess afterward is probably still easier than a heavy bureaucratic process where everything has to be rubber-stamped even among maintainers. :)

PR #153, opened 2023-04-23
ngfd: Bump SRCREV.

MagneFire:
Remove redundant patch file as the g_list_free_full change has been applied upstream now as well. Use a relative path to to the dbus-gmain.h file as the path to the dbus-gmain isn't actually passed as an include path by autotools. Reorder the inclusion of autotools and pkgconfig as autotools overrides the build directory which breaks the build as the compiled dbus-gmain library isn't in the expected location (alongside the sources).

Draft because this has only been tested on `hoki` which needs this update as it includes ffmemless support in the stock kernel as opposed to `bass` (custom kernel driver) or `sturgeon` (hal backend, like most others).


FlorentRevest (approved):

MagneFire:
Now tested on `sturgeon` works fine. Cannot test on another `ffmemless` device since I do not own a `bass`.

MagneFire:
Confirmed working on `dory` now too.

PR #163, opened 2023-05-19
Add patch to allow clean compile

beroset:
This adds a patch to add a missing include to allow rust to compile cleanly with newer versions of gcc.


MagneFire:
This adds three patch files but only uses one of them? Is this intentional?

beroset:
> This adds three patch files but only uses one of them? Is this intentional?
The first two patch files are from the original; I only added the third.  Weird `devtool` artifact?  I don't know.

beroset:
Revised and retested.  Yes, it was a devtool issue; I had used the `--force-patch-refresh` option by mistake.

FlorentRevest:
FWIW, I'm okay with carrying such a simple patch on our asteroid layer even though this has actually nothing to do with asteroid but we should make sure that the layer that provides rust-llvm is aware and/or contribute this patch there too, so that the maintenance of this doesn't belong to us.
Especially since we're not building this recipe as part of the image, I can imagine we'll just forget about this and it can remain there for a while :p 
Anyway, looking forward to see what sort of experiments you're doing with rust :p

PR #168, opened 2023-07-04
Factor out wait for /dev/.coldboot to separate command

beroset:
Every target except for the qemux86 target included android-init indirectly as part of the conf/machine/include/hybris-watch.inc file that each image includes.  Without this, asteroid-launcher for qemux86 never starts because it is waiting for something to create the file /dev/.coldboot_done.  Since every target except qemux86 already includes android-init, this does not affect any other target but fixes https://github.com/AsteroidOS/asteroid/issues/259 for qemux86.


MagneFire:
I'm fine with merging this. @dodoradio do you agree?

MagneFire (approved):

dodoradio:
I don't think it's a good idea to add a second init system to all builds just because hybris is our current default porting strategy.
Accepting this kind of android integration as default also breaks packaging for postmarketos, who won't ship any android dependency frameworks.
I think that asteroid-launcher should have a proper behaviour added instead of this plaster solution.

beroset:
@dodoradio has suggested that a better approach might be to conditionally add the check for `/dev/.coldboot_done` for watches that include `hybris-watch.inc` which I agree sounds like a better approach.  So I'm converting this to draft for now and will see about trying such an alternative.  Thanks all!

beroset:
I could create two different versions of `asteroid-launcher.service`, one of which omits this line: https://github.com/AsteroidOS/meta-asteroid/blob/8977f51d786121c20bc77c63b7b330d416d0145a/recipes-asteroid/asteroid-launcher/asteroid-launcher/asteroid-launcher.service#L10

Then we could selectively choose which one to install, based on whether the target it `qemux86` or not.  Is that a better way to do this?  Seems hackish, but I don't know how else to do it.

FlorentRevest:
I agree with @dodoradio here :) 

> I could create two different versions of `asteroid-launcher.service`, one of which omits this line:

I'd approve that but I wonder if it'll be more maintainable to have a `ExecStartPre=/usr/bin/asteroid-launcher-precondition` and have a script there that waits for `/dev/.coldboot_done` only if there is an android init installed for example. Or something along that line.

beroset:
This version was tested on both `catfish` and `qemux86` and works as expected.

recipes-asteroid/asteroid-launcher/asteroid-launcher/asteroid-launcher.service
```diff
@@ -7,7 +7,7 @@ ConditionUser=!root
 Type=notify
 WorkingDirectory=/
 EnvironmentFile=-/var/lib/environment/compositor/*.conf
-ExecStartPre=/bin/sh -ec 'while [ ! -f /dev/.coldboot_done ]; do sleep 1; done'
+ExecStartPre=/usr/bin/asteroid-launcher-precondition
 ExecStartPre=-/bin/sh -ec '/bin/echo QUIT > /run/psplash_fifo'
```
MagneFire, line 11: What's your opinion on moving this to the new script as well?

MagneFire (approved):
Minor comment, otherwise looks fine. Thank you :smile:

FlorentRevest:
Thank you!

PR #172, opened 2023-08-02
brcm-patchram-plus: Remove default service.

MagneFire:
The default service doesn't offer anything, it isn't useful on its own. Remove it such that build issues may become clear sooner.

This should fix an issue where the default patchram service is installed on the current nightlies.

This fixes: https://github.com/AsteroidOS/meta-smartwatch/issues/204


eLtMosen (approved):
Have now finally tested and can confirm its working!

PR #173, opened 2023-08-04
Add wifi-enabler recipe.

MagneFire:
Some Qualcomm based watches need additional steps to enable WiFi. Ideally this service should be added to the Android init.rc but this does not appear to work since the commands are executed too soon in the boot process.

This provides a systemd service such that WiFi is enabled later in the boot process.

### DIFF
--- a/recipes-android/wifi-enabler/wifi-enabler/wifi-enabler.service
+++ b/recipes-android/wifi-enabler/wifi-enabler/wifi-enabler.service
@@ -1,0 +1,11 @@
+[Unit]
+Description=Load the WiFi firmware and enable driver
+
+[Service]
+Type=simple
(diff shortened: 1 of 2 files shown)


recipes-android/wifi-enabler/wifi-enabler/wifi-enabler.service
```diff
@@ -0,0 +1,11 @@
+[Unit]
+Description=Load the WiFi firmware and enable driver
+
+[Service]
+Type=simple
+ExecStartPre=-/bin/sh -c "/bin/echo > /dev/wcnss_wlan"
```
FlorentRevest, line 6: Given that this is a Qualcomm specific driver, I think this service would be named in a way that's more specific like msm-wifi-enabler ? 
Otherwise lgtm

PR #154, opened 2023-04-28
Update to mickledore

beroset:
This is part of the fix for https://github.com/AsteroidOS/asteroid/issues/245 to update from kirkstone to mickledore.

PR #174, opened 2023-08-04
build ext4 images without ext4's metadata_csum feature

argosphil:
The "metadata_csum" feature changed in an incompatible fashion since the Linux 3.* kernel we use on `sparrow`, for example. This means that loop-mounting the ext4 on a modern Linux system and modifying it will result in a filesystem image that cannot be mounted by `sparrow`'s kernel.

The error message in question is:
```
[  156.079616](CPU:0-pid:235:dmesg) JBD2: Unrecognised features on journal
[  156.079639](CPU:0-pid:235:dmesg) EXT4-fs (mmcblk0p29): error loading journal
```

Marked as draft until I've actually built and tested this on `sparrow`. (I've tested on `beluga` and the `tune2fs -l` output looks okay...)

### DIFF
--- a/classes/asteroid-image.bbclass
+++ b/classes/asteroid-image.bbclass
@@ -21,3 +21,5 @@ EXTRA_USERS_PARAMS = "groupadd system; \
 
 IMAGE_OVERHEAD_FACTOR = "1.0"
 IMAGE_ROOTFS_EXTRA_SPACE = "131072"
+
+EXTRA_IMAGECMD:ext4 += " -O^metadata_csum"


beroset:
As part of my efforts to move from `kirkstone` to `mickledore`, I made a similar change here:  https://github.com/AsteroidOS/meta-asteroid/pull/154/files#diff-3f345f9918c6cf6fde6a897bb7823dc2e073a3904faa5174c14082f9d5006c4b

argosphil:
@beroset Thanks!

I'm merely curious, but how does this relate to the `mickledore` update?

beroset:
The relationship is that with `mickledore`, the flag was on by default and so has to be explicitly turned off to allow the watch to boot.  Otherwise the watch reports errors like this when trying to mount the image:
```
[  487.166959] EXT4-fs (mmcblk0p35): couldn't mount as ext3 due to feature incompatibilities
[  487.168302] EXT4-fs (mmcblk0p35): couldn't mount as ext2 due to feature incompatibilities
[  487.169564] EXT4-fs (mmcblk0p35): Couldn't mount because of unsupported optional features (2000)
```

argosphil:
Thanks again.

The flag is on in `kirkstone` as well; it's just that as the ext4 fs isn't ever modified after creation, the journal is empty, so it's not recognized as incompatible. My guess is `mickledore` generates file system images differently?

argosphil:
Built and tested on `sparrow`; works as intended.

FlorentRevest:
Great detective work, thank you for your debugging! ;)

PR #178, opened 2023-08-27
nfcd: Don't autostart service.

MagneFire:
NFC seems to cause issues on multiple watches.
On `beluga` it causes for a screen flickering issues on the sides.
On `ray` it causes for random horizontal lines to appear as well as ghost touchscreen presses.

As there's currently little use of NFC disable the service from starting to workaround these issues.

This cause of this issue was first discovered by @argosphil.


FlorentRevest:
> On `beluga` it causes for a screen flickering issues on the sides.
> On `ray` it causes for random horizontal lines to appear as well as ghost touchscreen presses.

Woah! :smile: Funky ahah. There is gonna be some fun root-causing to figure this out the day when we want to enable it again.

But anyway, in the meantime, that change is perfectly fine by me. I think we should even drop nfcd from our images entirely since we don't use it anywhere and afaik no one ever looked into this. It could be a really cool addition but no point having a half baked support.

PR #179, opened 2023-11-25
enable restart-on-failure for bluetooth service

dodoradio:
Some watches suffer from occasional bluez crashes for various reasons, likely due to issues with the vendor bluetooth blobs. Bluetooth sync is important for the asteroid experience, this change enables restart on failure so that phone synchronisation is a bit more stable.


beroset (approved):
Looks OK to me.  For watches on which BlueZ doesn't fail, this will have no effect and for the others it may "fix" a problem.

FlorentRevest:
FWIW the commit introducing these comments for convenience doesn't warn against any nefarious side-effect from these either ;)

PR #180, opened 2023-11-25
Patch qtwayland to fix closing apps.

dodoradio:
This stops qtwayland destroying the client immediately after requesting a graceful close. Destroying the wayland client would result in applications always crashing. This patch inelegantly stops this behaviour.
This should hopefully finally fix https://github.com/AsteroidOS/asteroid-launcher/issues/163

### DIFF
--- a/recipes-qt/qt5/qtwayland/0004-Don-t-destroy-wayland-client-after-requesting-a-grac.patch
+++ b/recipes-qt/qt5/qtwayland/0004-Don-t-destroy-wayland-client-after-requesting-a-grac.patch
@@ -1,0 +1,28 @@
+From 1a3fafbe7c9c384e7dd7f84e3bcf3de649c576f0 Mon Sep 17 00:00:00 2001
+From: Arseniy Movshev <dodoradio@outlook.com>
+Date: Sat, 25 Nov 2023 15:33:21 +0000
+Subject: [PATCH] Don't destroy wayland client after requesting a graceful
+ close from the same client.
+
+---
+ src/compositor/compositor_api/qwaylandcompositor.cpp | 4 ++--
+ 1 file changed, 2 insertions(+), 2 deletions(-)
+
+diff --git a/src/compositor/compositor_api/qwaylandcompositor.cpp b/src/compositor/compositor_api/qwaylandcompositor.cpp
+index 98a0bb7f..e75e4b65 100644
+--- a/src/compositor/compositor_api/qwaylandcompositor.cpp
++++ b/src/compositor/compositor_api/qwaylandcompositor.cpp
+@@ -749,8 +749,8 @@ void QWaylandCompositor::destroyClient(QWaylandClient *client)
+     QWaylandQtWindowManager *wmExtension = QWaylandQtWindowManager::findIn(this);
+     if (wmExtension)
+         wmExtension->sendQuitMessage(client);
+-
+-    wl_client_destroy(client->client());
++    else
++        wl_client_destroy(client->client());
+ }
+ 
+ /*!
+-- 
+2.40.1
+
--- a/recipes-qt/qt5/qtwayland_git.bbappend
+++ b/recipes-qt/qt5/qtwayland_git.bbappend
@@ -1,6 +1,7 @@
 FILESEXTRAPATHS:prepend := "${THISDIR}/qtwayland:"
 SRC_URI += " file://0001-Forces-GLES2-the-dirty-way.patch \
              file://0002-Revert-most-of-Remove-QWaylandExtendedSurface-from-t.patch \
-             file://0003-xdg-shell-Add-support-for-extended-surfaces.patch"
+             file://0003-xdg-shell-Add-support-for-extended-surfaces.patch \
+             file://0004-Don-t-destroy-wayland-client-after-requesting-a-grac.patch"
 
 DEPENDS:remove = "${XKB_DEPENDS}"


dodoradio:
looks like some changes to asteroid-launcher will also be needed here. Currently the launcher is written with the assumption that applications immediately disappear as soon as `closeClientForWindowId` is called. If applications take some time to close (which is the case with this PR) then the launcher need some additional logic to handle applications which are in the process of closing themselves.

recipes-qt/qt5/qtwayland/0004-Don-t-destroy-wayland-client-after-requesting-a-grac.patch
```diff
@@ -0,0 +1,28 @@
...
+--- a/src/compositor/compositor_api/qwaylandcompositor.cpp
++++ b/src/compositor/compositor_api/qwaylandcompositor.cpp
+@@ -749,8 +749,8 @@ void QWaylandCompositor::destroyClient(QWaylandClient *client)
+     QWaylandQtWindowManager *wmExtension = QWaylandQtWindowManager::findIn(this);
+     if (wmExtension)
+         wmExtension->sendQuitMessage(client);
+-
+-    wl_client_destroy(client->client());
++    else
++        wl_client_destroy(client->client());
```
FlorentRevest, line 22: Diving in the implementation of `wl_client_destroy` https://gitlab.freedesktop.org/wayland/wayland/-/blob/main/src/wayland-server.c#L921 it appears that skipping it is a logical flaw. That's actually a memory leak (keeping some file descriptors open, like the client->connection and dynamic memory allocated like the client pointer) 

This "appears" to be a fix for the problem you've noticed just because it keeps dangling references alive but I don't think that's a fix, it's just hiding a visible bug (sigsegv with logs) with an invisible bug (memory leak) 

I see that newer versions of QtWaylandCompositor are still operating the same way https://code.qt.io/cgit/qt/qtwayland.git/tree/src/compositor/compositor_api/qwaylandcompositor.cpp?h=dev#n731 and I believe the problem is actually on the client side not handling the client_destroy properly.

I appreciate the time you've spent diving in this issue and I sympathize with the cause of removing these ugly SIGSEGV from our logs but I don't think this fix is correct. I wouldn't be surprised if an update of qtwayland would automagically solve this bug anyway. We're currently lagging quite a bit behind and I fear that a fix that isn't one would just add tech-debt on our migration path.

dodoradio: I definitely didn't go deep enough to notice the memory leak, thank you for that. 
As far as I've understood, the quit request is non-blocking. this means that applications do get killed, and never have a chance to exit properly. 
The memory leak is a good reason to block this PR, but the segfaults in logs are bigger than just looking ugly. for example, they result in GPS never being shut down and draining the battery quickly, and we can't actually have any `onAboutToBeDestroyed` statements in our qml because they'd never be run.

dodoradio: I've not seen this chunk of code touched in ths latest qtwayland. unless that quit request has been changed to block until the application exits, applications will continue to crash even under newer versions of qtwayland. I'll have to go back into this code to see if some wait can be put in place until the application has had time to close itself.

FlorentRevest: I think the compositor is a red herring. There is no reason why closing the Wayland connection should crash an app. The compositor should be free to close this socket. If the clients segfault as a result, then the but is in the client. Any solution in compositor code will be a hack to workaround client bugs. If I were you I'd rather collect a backtrace from the client at the time of the SIGSEGV and figure out why it's crashing on connection close. This is likely to be a bug in qtwayland but in the client version of this.

If I had to give an analogy, this would be like Firefox crashing if it connects to a website that 404 and then decide to change apache to replace 404s with 405s as a workaround to avoid Firefox crashes, sure enough it would make the life of Firefox users nicer but this would overall be an inappropriate fix. There's no reason why a client should crash on connections close.

dodoradio: sorry, I don't know why the conversation of segfaults came up, not sure why I repeated 'segfault'. the issue is that the app 'willingly' closes itself with the log message `did the compositor die?` - but, since this is an error condition, it doesn't do any cleanup. it's not a crash, it's an error handler that behaves like a crash, so the clients are working as intended

dodoradio: `May 01 19:43:49 koi asteroid-launcher[471]: The Wayland connection broke. Did the Wayland compositor die?`

FlorentRevest: The point remains that there's no reason why a process should emergency exit without cleanups if a connection it's monitoring is closed. If we need to dive in a direction, I'd rather dive into the code path that leads to this rather odd decision of exiting without cleanups

dodoradio: I think it's quite reasonable that a gui application does not exit gracefully if it suddenly notices it's no longer a gui application. I don't think it should have any obligation to finish running any more UI code (eg. the qml that triggers GPS) if it's not actually a UI application
I umderstand from my reading of the wayland spec that you're not meant to just yank away the wayland connection like this. the application needs to be the one initiating the closing of its display connection (since in the simplest wayland there's no shell and the application provides its own window decoration) and if it's forcibly removed (not even somehow 'gracefully' closed, just destroyed as an object) it's reasonable for the application to assume that the compositor has catastrophically failed and that it needs to exit asap so it does not get lost in the inevitable tumbleweed of other crashing applications

FlorentRevest: Dying without cleanups is only appropriate for a client if the compositor does something outrageously unexpected and out of spec. Before blaming qtwayland for this, which was extensively tested, I'd rather blame our unmaintained lipstick fork instead.

The way we "close apps" is by closing the wl_client object https://github.com/AsteroidOS/lipstick/blob/master/src/compositor/lipstickcompositor.cpp#L279 if I were you I'd play with variations of this like destroying the surface or closing the  extended surface instead of the client connection. This could be more fruitful.

dodoradio: `Dying without cleanups is only appropriate for a client if the compositor does something outrageously unexpected and out of spec` The application does say 'did the compositor die?' the compositor is only meant to ever break a connection like this if it has completely broken catastrophically. The wayland shell extension is the intended method for getting applications to close.
I talked to notkit, who said that internally sailfish uses `destroysurface`. This works, but it's a hacky workaround, not the more correct way of doing it. https://doc.qt.io/qt-6/qwaylandclient.html#close is brief, but the fact that it internally calls the xdg quit suggests that it is the intended mechanism here.  https://wayland.app/protocols/xdg-shell#xdg_toplevel:event:close is absolutely the correct way of doing this, and anything other than close() is a hacky workaround. 
I'm also just not sure qtwayland is actually used enough for this sort of bug to be noticed and fixed.

PR #184, opened 2024-02-08
Disable DeviceId in setttings to be able provide own implementation

jmlich:
see https://github.com/AsteroidOS/asteroid-btsyncd/pull/31


FlorentRevest:
This PR looks good to me but should probably be submitted in sync with the btsyncd code.

PR #185, opened 2024-02-08
[nemo-qml-plugin-systemsettings] Parse manufacturer and device name

jmlich:
SailfishOS relies on `/etc/hw-release` which doesn't exists in Asteroid OS.

The values are taken from /system/build.prop if wasn't loaded before.

This is related to https://github.com/AsteroidOS/asteroid-btsyncd/pull/31 and https://github.com/AsteroidOS/asteroid-settings/pull/97

Please note, that I have very few experience with bitbake. This patch was crafted manually and need careful review and test.

### DIFF
--- a/recipes-nemomobile/nemo-qml-plugins/nemo-qml-plugin-systemsettings/0007-Parse-manufacturer-and-device-name-from-system-build.patch
+++ b/recipes-nemomobile/nemo-qml-plugins/nemo-qml-plugin-systemsettings/0007-Parse-manufacturer-and-device-name-from-system-build.patch
@@ -1,0 +1,66 @@
+From e3172aed8d0d68b9aa79a1a4a942c80ec298dc8c Mon Sep 17 00:00:00 2001
+From: Jozef Mlich <jozef.mlich@greycortex.com>
+Date: Thu, 8 Feb 2024 17:27:26 +0100
+Subject: [PATCH] Parse manufacturer and device name from /system/build.prop if
+ wasn't loaded before
+
+---
+ src/aboutsettings.cpp | 15 ++++++++++++++-
+ src/aboutsettings_p.h |  1 +
+ 2 files changed, 15 insertions(+), 1 deletion(-)
+
+diff --git a/src/aboutsettings.cpp b/src/aboutsettings.cpp
+index 414e6ed..c58232e 100644
+--- a/src/aboutsettings.cpp
++++ b/src/aboutsettings.cpp
+@@ -72,6 +72,10 @@ void parseReleaseFile(const QString &filename, QMap<QString, QString> *result)
+                 continue;
+             }
+ 
++            if (line == "") {
++                continue;
++            }
++
+             QString key = line.section('=', 0, 0);
+             QString value = line.section('=', 1);
+ 
+@@ -86,7 +90,7 @@ void parseReleaseFile(const QString &filename, QMap<QString, QString> *result)
+             //
+             // see http://stackoverflow.com/a/2821183
+             // and http://stackoverflow.com/a/2821201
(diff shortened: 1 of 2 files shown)


FlorentRevest:
The problem with `/system/build.prop` is that this is an android file and not all our ports are based on an Android HAL. Also, even  some Android ports could be missing this file or have incorrect values there for one reason or another. (sometimes we could use the /system of another watch or whatever). Also we lack control on the strings written there (for example, maybe we want to write LG rather than LGE). This is why SailfishOS implemented the "hw-release" file where they can define an arbitrary hardware name, even on devices that don't have an android `/system/build.prop` and also add other values.

This is why I feel we should start providing `hw-release` files for our watches rather than growing a dependency to android stuff.

In AsteroidOS, we have a vaguely similar concept of "device info" in `/etc/asteroid/machine.conf`. We use it for device specific information such as "does it have a round screen". Seeing how it's implemented should give you an idea of how to generate `hw-release` files.

- each port has a layer in meta-smartwatch and a MACHINE definition. MACHINE specific values can be defined, for example, bass defines `MACHINE_DISPLAY_ROUND = "true"` https://github.com/AsteroidOS/meta-smartwatch/blob/master/meta-bass/conf/machine/bass.conf#L9
- all AsteroidOS builds build the "asteroid-machine-config" recipe which uses these macros to conditionally generate a MACHINE specific `/etc/asteroid/machine.conf` file: https://github.com/AsteroidOS/meta-asteroid/blob/master/recipes-asteroid/asteroid-machine-config/asteroid-machine-config_1.0.bb this is a bit of clever python but at the end of the day it's quite straightforward, it reads the MACHINE_DISPLAY_ROUND value set by the machine from the build environment and generates a file for it.

You could introduce a similar mechanism to generate hw-release files. You'd first need to modify all `meta-smartwatch/meta-*/conf/machine/*.conf` files and add macros such as `MACHINE_VENDOR` and `MACHINE_NAME` there. Then you could modify asteroid-machine-config to generate `/etc/hw-release` (with some fallback strings if we build on a machine that doesn't have this macro defined. And then everything would work ;)

Thank you!

beroset:
For a number of AsteroidOS images, there is more than one manufacturer and model, and there's simply no way to supply that information when compiling because the *only* difference is the case in which the watch is packaged.  See https://asteroidos.org/watches/firefish/ as an example: `ray` shows 11 different models and 7 different manufacturers.

FlorentRevest:
> For a number of AsteroidOS images, there is more than one manufacturer and model, and there's simply no way to supply that information when compiling because the _only_ difference is the case in which the watch is packaged. See https://asteroidos.org/watches/firefish/ as an example: `ray` shows 11 different models and 7 different manufacturers.

Indeed. But on the other hand, we also tend to ship a custom /system/ directory with a custom build.prop there so the manufacturer and model will also be wrong there :/

FlorentRevest:
We stopped shipping nemo-qml-plugin-systemsettings so I'll close this as stale

jmlich:
So what is recommended replacement?

FlorentRevest:
Now the equivalent code is located in qml-asteroid. I'm sorry your series of PRs from way back has been stuck in limbo over the years. The cmake stuff on the phone side seems very useful but I think there's just no one using this code so it's totally unmaintained and we can't really test it either, that'd explain why those telescope changes didn't receive any love. As for the changes on the watch side, they were scattered in various places and I never got a feeling I really understood what this was about so I think it drowned over time. I'd understand if you've already been frustrated enough contributing here but if you're up to refresh those changes, I promise to have time for quick reviews in the next two weeks. If it no longer makes sense or is no longer useful, let's close the stale PRs and issues. 
Thank you!

PR #187, opened 2024-02-11
Remove unconditional hybris dependencies

casept:
Some components in AsteroidOS unconditionally depend on libhybris, and many default configs assume a hybris-based device. This PR breaks out these dependencies, paving the way for easily supporting devices running mainline without Android drivers.

Changes to meta-smartwatch are not required. 

I have tested these changes on catfish and they seem to work fine there, but more testing on other watches is appreciated, given how broad the changes are.

### DIFF
--- a/conf/machine/include/hybris-watch.inc
+++ b/conf/machine/include/hybris-watch.inc
@@ -1,4 +1,5 @@
 MACHINE_FEATURES = "alsa bluetooth usbgadget usbhost ext2"
+MACHINEOVERRIDES =. "hybris-machine:"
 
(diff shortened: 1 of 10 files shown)


FlorentRevest:
Looking at it more closely a second time, I think you were over-zealous in your sensorfw patches strategy, all the patch should apply to both mainline and hybris machines. Only the primaryconfig should be different on one or the other

casept:
As requested, the sensorfw patches are now applied on all machines and the qtsensors patch has been dropped.

FlorentRevest:
Suggested rather than requested 😄 but that looks perfect now, thank you for your patience!

PR #190, opened 2024-03-02
Fix pulseaudio modules

### LINKED ISSUE #189: Sound is not working
As several people have noted in the matrix chat, sound is not working.  The symptom is that if you go to Settings->Sound and alter the volume, then leave that screen and come back, the volume is reset to 0% no matter what was previously set.  Additionally, sound is not working - that is, there is no audio output through the speaker.

beroset:
This fixes #189 by altering how the pulseaudio modules are built and installed.  Since pulseaudio 16.0, modules no longer use the version number as part of the installation path.  That is, files that were once in /usr/lib/pulse-15.0/modules are now simply in  /usr/lib/pulseaudio/modules

See https://www.freedesktop.org/wiki/Software/PulseAudio/Notes/16.0/#moduleinstallationlocationchangedremembertoupgradepaprefstothelatestversion


FlorentRevest:
Neat! Thank you :)

PR #193, opened 2024-05-25
Fix pseudo-native to compile cleanly with gcc14

beroset:
This applies some fixes to allow gcc14 to cleanly compile pseudo-native. It partially addresses [#287](https://github.com/AsteroidOS/asteroid/issues/287).


FlorentRevest:
This looks good to me but I wonder how much of a rabbithole these gcc14 bugs are ? It seems like OE should be the one dealing with this and if we're lagging behind due to an old OE version we should rather update our base OE version

beroset:
Indeed, OE (and Yocto) developers *are* dealing with it, which is why this is still a draft version.  It might take a while for all of these fixes to get into the mainline, but I'm trying to help that along for both [`pseudo`](https://git.yoctoproject.org/pseudo/commit/?id=e11ae91da7d0711f5e33ea9dfbf1875dde3c1734) and also [`cracklib`](https://github.com/cracklib/cracklib/issues/74).  The longer term fix is likely going to be exactly as you suggest, which is to update our versions.

beroset:
These issues are now all addressed, largely via upstream patches.

PR #199, opened 2024-09-08
qt5-qpa-hwcomposer-plugin: Bump to fix issues with hwc2

MagneFire:
Fixes issue with loading hwc2 in case the headers exist but the device doesn't support it. Also cleans-up Always-on-Display patch to use the base power_module.

Also adds Qt6 support.

This fixes hwcomposer issues on `sawfish` (oreo and pie base).
Needs more testing on other platforms before it can be merged.


FlorentRevest (approved):
lgtm at least :) thank you for this!

MagneFire:
With the move to Qt6 this PR is no longer relevant.

PR #200, opened 2024-10-06
swclock-offset: Bump SRCREV

MagneFire:
Upstream now also uses /usr. So we don't need to manually state that it needs to be installed there.
Additionally OpenRC and systemd services files are always installed, so we need to remove the OpenRC files manually to fix packaging issues.

This adjusts the recipe to work with these MRs:
- https://gitlab.com/postmarketOS/swclock-offset/-/merge_requests/4
- https://gitlab.com/postmarketOS/swclock-offset/-/merge_requests/5


MagneFire:
The project appears to be archived upstream, I'll try to figure out why...

MagneFire:
Thanks for the quick merge though :D

MagneFire:
Okay, seems to be fine for now. They are migrating to a selfhosted instance. Which just started.

https://status.postmarketos.org/

image

More information: https://gitlab.com/postmarketOS/postmarketos/-/issues/77

PR #201, opened 2024-10-18
usb-moded: Create flag file /etc/usb-debugging-enabled

MagneFire:
Upstream changed the ConditionPathExists to use a different path as it's more suitable for build time files. [^1]

Adjust the recipe to create the file on this location similarly to how upstream achieves this. [^2]

Draft PR because this still needs to be tested.

This potentially fixes: https://github.com/AsteroidOS/meta-smartwatch/issues/253

[^1]: https://github.com/openembedded/meta-openembedded/commit/3309a9d4dde1b8cbab60322ec24e121330e23cea
[^2]: https://github.com/openembedded/meta-openembedded/commit/145ae5af9ef45f566f07751a2c2ea6c224e71a24

### DIFF
--- a/recipes-nemomobile/usb-moded/usb-moded_git.bb
+++ b/recipes-nemomobile/usb-moded/usb-moded_git.bb
@@ -74,7 +74,7 @@ do_install:append() {
     install -d ${D}/var/lib/misc/
     touch ${D}/var/lib/misc/udhcpd.leases
 
-    touch ${D}/var/usb-debugging-enabled
+    touch ${D}/etc/usb-debugging-enabled
 }
 
 FILES:${PN} += " ${systemd_system_unitdir} /usr/share/dbus-1/services/ /var/lib/misc/udhcpd.leases"


beroset:
I did a fresh build of `catfish` and flashed that image and then verified that ADB did not work.  Then I applied this patch, rebuilt, reflashed and indeed, now ADB works.  Trying `dory` next.

beroset:
Same test with `dory` and same result.  Looks good!

beroset:
Also just tested `sturgeon` and same happy result.

beroset (approved):
Tested on `catfish`, `dory` and `sturgeon` and was able to verify that ADB did not work before applying this, but does work after applying it.

MagneFire:
@beroset Thanks for testing!

Currently I'm trying to take advantage of `USB_DEBUGGING_ENABLED="1"` as that should also create the `/etc/usb-debugging-enabled` file, and it might be the better solution since we don't need to manually maintain it here.

But it does not appear to be created when analyzing the `ext4`. Here's what I've tried so far:
- Added that flag to: `build/conf/local.conf`, `meta-asteroid/classes/asteroid-image.bbclass`, `meta-asteroid/conf/distro/asteroid.conf`.
- Adjust the `https://github.com/openembedded/meta-openembedded/blob/scarthgap/meta-oe/recipes-devtools/android-tools/android-tools_5.1.1.r37.bb#L195`:
  - `ROOTFS_POSTPROCESS_COMMAND_${PN}-adbd += "android_tools_enable_devmode;"`
  - `ROOTFS_POSTPROCESS_COMMAND:append = " android_tools_enable_devmode;"`
  - `ROOTFS_POSTPROCESS_COMMAND:append = " ${@bb.utils.contains("USB_DEBUGGING_ENABLED", "1", "android_tools_enable_devmode;", "", d)}"`
- Complete rebuild.

I'll do some more testing, if you have some idea of what I could be doing incorrectly, then that would be welcome as well :D

beroset:
I tried a few things with `USB_DEBUGGING_ENABLED` but was also unsuccessful with all of my attempts.  Maybe we merge this and see if we can figure out something more elegant later?

MagneFire:
Sounds good, thanks for helping with this one!

PR #202, opened 2024-12-23
ngfd: Bump SRCREV

MagneFire:
This includes a crash fix for the ffmemless plugin(https://github.com/sailfishos/ngfd/commit/80d356b1c410954c756d3f2cc783a4608e0122b9).

This fixes ngfd from crashing when the alarm fires on `hoki`.


beroset:
Tested and verified.

PR #204, opened 2025-02-11
Several fixes for build errors as well as refactoring of how users and groups are handled

dv1:
These commits introduce fixes for build errors that happen when -Werror is used during compilation. Until now, some rather ugly workarounds were being used that were fragile, for example, by using sed to remove "-Werror" from build scripts. The fixes here patch the code that causes -Werror to trigger build errors, making such hacks unnecessary.

These commits also rework how users and groups are configured. They introduce a new bbclass where AsteroidOS specific users and groups are created centrally. This removes the need for the "host-user-contaminated" `INSANE_SKIP`, and also uses named constants instead of hardcoded UIDs for the ceres user and group.


beroset (approved):
Used and tested.  This is a valuable contribution, thank you!

PR #205, opened 2025-03-18
Fix geoclue postinstall

beroset:
This fixes the postinstall step required with geoclue by adding gsettings to the inherited classes.  That class automatically performs the step of compiling the schemas using glib-compile-schemas and also adds the /usr/share/glib-2.0/schemas to the FILES variable for the project, which is why it was removed from this file.

The gsettings also has a postinstrm step that recompiles the schemas after the package is removed.

This fixes https://github.com/AsteroidOS/asteroid/issues/304


MagneFire (approved):
This is awesome! Thanks!

PR #207, opened 2025-04-14
Add patch for python3-pyqt5 recipe

beroset:
This allows us to build the python3-pyqt5 recipe without having to add QtWidgets or mesa.  It also applies some patches from ST to fix yocto build errors.

This fixes https://github.com/AsteroidOS/asteroid/issues/227 and https://github.com/AsteroidOS/asteroid/issues/98


MagneFire:
Awesome thanks!

PR #220, opened 2026-02-05
Add support for Android boot image formats V1, 2, 3 and 4.

Aiden-Isik:
This PR adds support for the modern Android boot image formats, enabling devices which were shipped with Android 10 or above to be ported to AsteroidOS.

Changes:
- Updated the mkbootimg part of mkbootimg-tools to AOSP latest on the main branch.
- Rewrote the mkbootimg class to support all currently existing Android boot image formats, whilst retaining backwards compatibility for watches which used the old class (specifically tetra).


Aiden-Isik:
Closing as I am opening a new PR with an expanded scope which includes these changes.

PR #232, opened 2026-03-23
Remove compile-time dependency on virtualkeyboard

beroset:
This remove the compile-time dependency on qtvirtualkeyboard because it is not needed.  See also https://github.com/AsteroidOS/qml-asteroid/pull/88 for the corresponding change in the qml-asteroid build.


FlorentRevest:
I think we de discussed in the qml-asteroid change that we'll keep that logic for Qt6 so let's abandon this PR

PR #209, opened 2025-10-02
Upgrade nemomobile dependencies

casept:
Needs to be done in preparation for upgrading to Qt6, as upstream has recently started adding support for it.

The plan is to go through recipes one-by-one and sync with upstream as I have time for. Help is welcome.

PR #234, opened 2026-04-09
Fix build on modern Linux distros (GCC 15, glibc 2.43)

KonTy:
Title: Fix build on modern Linux distros (GCC 15, glibc 2.43)

---

Hi,

I have been trying to build AsteroidOS from source for a while now and I kept hitting build failures. It turns out that AsteroidOS does not compile at all on any recent Linux distribution. I am on Arch which ships GCC 15 and glibc 2.43, but anyone on Fedora 42 or even Ubuntu 25.10 will hit the same wall. These are not edge cases, these are the defaults now on most distros people actually use for development.

I spent a good amount of time tracking down what was going wrong and it came down to four separate issues. I fixed all of them and tested everything thoroughly before submitting this.

Here is what was broken:

1) dtc would not compile because GCC 15 made some warnings into errors. Two variables had wrong const qualifiers. Straightforward fix, works fine on older compilers too. I ran the full dtc test suite after patching it, all 8 test suites passed.

2) nfcd had a race condition in its build system. When you build with parallel make (which everyone does), the debug and release builds both try to generate code in the same directory at the same time. One process would overwrite a header file while the other was still reading it, so you would get random compile errors about unknown types. I added a dependency so they run one after the other. Simple and reliable.

3) mesa would not build because glibc 2.43 now defines C11 thread types like once_flag directly in its headers. Mesa has its own bundled copy of those same types and they clash. Mesa actually already had code to handle this but it was only enabled for Android builds for some reason. I removed that restriction so it works on Linux too. I also had to clean up some EGL code that was mixing C11 and POSIX thread types which are now different structs in glibc.

4) virglrenderer had the exact same glibc conflict as mesa. Same root cause, similar fix. I had to add the detection logic, update the config template so the build system actually passes the flag through, and add a missing include that used to work by accident.

Everything is done through bbappend files so I did not touch any upstream OE recipes. Each fix is in its own commit.

For testing I did a complete asteroid-image build for the emulator target. All 8101 build tasks completed successfully. I also ran the upstream test suites, dtc passed all 8 out of 8 test suites and virglrenderer passed 7 out of 7 tests (the 8th is a hardware GPU test that is expected to time out on a build machine without a real GPU).

I think this is pretty important to get in. Right now if someone follows the build instructions on a modern system they will just get a wall of errors and no working image. That is probably turning people away. These fixes are small and safe and do not change anything for people on older toolchains.

Let me know if you want me to change anything or split this up differently.

### DIFF
--- a/recipes-graphics/mesa/mesa.bbappend
+++ b/recipes-graphics/mesa/mesa.bbappend
@@ -1,0 +1,3 @@
+FILESEXTRAPATHS:prepend := "${THISDIR}/${PN}:"
+
+SRC_URI += "file://0001-meson-use-system-C11-threads-when-available.patch"
--- a/recipes-graphics/mesa/mesa/0001-meson-use-system-C11-threads-when-available.patch
+++ b/recipes-graphics/mesa/mesa/0001-meson-use-system-C11-threads-when-available.patch
@@ -1,0 +1,106 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: konty <konty@users.noreply.github.com>
+Date: Wed, 09 Apr 2026 00:00:00 +0000
+Subject: [PATCH] Fix build with glibc 2.43+ system C11 threads
+
+glibc 2.43+ exposes once_flag/call_once from <stdlib.h>, conflicting
+with mesa's bundled C11 threads that redefines once_flag as
+pthread_once_t. Remove the Android-only HAVE_THRD_CREATE guard so
+mesa uses the system <threads.h> on any platform that provides it.
+
+Additionally, the EGL sync code mixed C11 types (cnd_t/mtx_t) with
+POSIX calls (pthread_cond_init with CLOCK_MONOTONIC condattr). With
+system <threads.h>, cnd_t and pthread_cond_t are distinct union types
+in glibc 2.43+. Switch dri2_egl_sync to use pthread types consistently.
+
+Upstream-Status: Backport [Mesa 24.1+]
+Signed-off-by: konty <konty@users.noreply.github.com>
+---
+--- a/meson.build	2024-05-08 06:28:59.000000000 -0700
++++ b/meson.build	2026-04-09 07:21:41.472965796 -0700
+@@ -1423,11 +1423,9 @@
+ 
+ with_c11_threads = false
+ if cc.has_function('thrd_create', prefix: '#include <threads.h>')
+-  if with_platform_android
+-    # Current only Android's c11 <threads.h> are verified
+-    pre_args += '-DHAVE_THRD_CREATE'
+-    with_c11_threads = true
+-  endif
++  # Use the system C11 threads when available (required for glibc 2.43+)
++  pre_args += '-DHAVE_THRD_CREATE'
++  with_c11_threads = true
+ endif
+ 
+ if cc.has_header_symbol('errno.h', 'program_invocation_name',
+--- a/src/egl/drivers/dri2/egl_dri2.h	2024-05-08 06:28:59.000000000 -0700
++++ b/src/egl/drivers/dri2/egl_dri2.h	2026-04-09 07:22:10.320720748 -0700
+@@ -428,8 +428,8 @@
+ 
+ struct dri2_egl_sync {
+    _EGLSync base;
+-   mtx_t mutex;
+-   cnd_t cond;
++   pthread_mutex_t mutex;
++   pthread_cond_t cond;
+    int refcount;
+    void *fence;
+ };
+--- a/src/egl/drivers/dri2/egl_dri2.c	2024-05-08 06:28:59.000000000 -0700
++++ b/src/egl/drivers/dri2/egl_dri2.c	2026-04-09 07:22:10.336522656 -0700
+@@ -3365,7 +3365,7 @@
+    if (p_atomic_dec_zero(&dri2_sync->refcount)) {
+       switch (dri2_sync->base.Type) {
+       case EGL_SYNC_REUSABLE_KHR:
+-         cnd_destroy(&dri2_sync->cond);
++         pthread_cond_destroy(&dri2_sync->cond);
+          break;
+       case EGL_SYNC_NATIVE_FENCE_ANDROID:
+          if (dri2_sync->base.SyncFd != EGL_NO_NATIVE_FENCE_FD_ANDROID)
+@@ -3497,7 +3497,7 @@
+        dri2_sync->base.SyncStatus == EGL_UNSIGNALED_KHR) {
+       dri2_sync->base.SyncStatus = EGL_SIGNALED_KHR;
+       /* unblock all threads currently blocked by sync */
+-      err = cnd_broadcast(&dri2_sync->cond);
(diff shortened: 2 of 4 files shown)


FlorentRevest:
Thank you KonTy!

I feel that the right place for the mesa, virglrenderer and dtc fixes would be upstream yocto repositories. Did you send the fixes here because you think they wouldn't accept those on the walnascar branch we use ? Do you have any clue whether they could they take those changes on the whinlatter branches ? Curiously enough, we don't have much experience contributing this kind of build fixes to upstream Yocto so I'm not too familiar with the process.

If these fixes can't be contributed to the Yocto branches we use, then we could take the bbappend in meta-asteroid but I fear this is onboarding a bit of tech debt for us.

---

Re nfcd, the base recipe is being upgraded by https://github.com/AsteroidOS/meta-asteroid/pull/209. Since this is a super massive change and its author has been working on it for a long time and it is seemingly approaching its final state, I don't want to throw more rebase burden on its author. If that's ok with you, we'll first work on getting that nemomobile update PR in before getting back to the PR here.

Also, the right place to fix this would be https://github.com/sailfishos/nfcd. I expect they would be open to a PR for your patch and this would also keep our maintenance burden more sustainable.

KonTy:
@FlorentRevest honestly I'm just learning the codebase, and I'm trying ti figure out what's what, let me see if I can figure out those dependencies. Sorry about that. I just got the watch earlier and very excited about your project. I want to help out, I'm a developer of Microcore (still private since it is not done yet but I can make you contributes so you can help me test and give me feature requests :) ) which is a fitness app, completely open source and local no cloud needed you own your data. So I want to try to integrate the app and the watch, I think it would be so cool

Please bear with me, don't kick me too hard, I'm just learning my way around. I need your help to understand the system, and I want to become a contributor eventually.

MagneFire:
From a quick look at least the dtc fix seems to be available upstream: https://github.com/openembedded/openembedded-core/commit/28552a7b6c94060c7ab3899619ab8afb74124d02

beroset:
I've actually been on Fedora 43 for a while now and have been using gcc15 since Fedora 42, and AsteroidOS compiles for me without error.  I'm curious about what the difference is.  For the first check, I looked at the source code of `dtc` and found that the changes you propose are already in the code.  Since that recipe is from `oe-core`, can you please go into the `asteroid/src/oe-core` directory and make sure you're on the `walnascar` branch?  When I go in there and do `git log --oneline`, I get this:
```
bba638620f (HEAD -> walnascar, origin/walnascar) cve-check: fix debug message
ff1c54df4e (tag: yocto-5.2.4, tag: 2025-04.4-walnascar, tag: 2025-04.4) build-appliance-image: Update to walnascar head revision
c171a41e58 tiff: patch CVE-2025-8961
af4a1f0140 tiff: patch CVE-2025-9165
17a71c67a8 tiff: ignore CVE-2025-8851
09a9653c51 cups: patch CVE-2025-58364
cd732eb0cf cups: patch CVE-2025-58060
...
```
Does that match yours?

KonTy:
Thank you guys for your great advise I will look at it later today when I have time

KonTy:
I've updated this PR to drop the \`dtc\` and \`nfcd\` workarounds as requested. 

1. **\`dtc\`**: Dropped. Confirmed it builds cleanly without changes on OE-core's \`walnascar\` branch.
2. **\`nfcd\`**: Dropped from \`meta-asteroid\`. I have successfully isolated this race condition onto the \`nfcd\` \`src/Makefile\` specifically and submitted an upstream PR here: https://github.com/sailfishos/nfcd/pull/27

Regarding \`virglrenderer\` and \`mesa\` glibc 2.43 C11 threading issues: 
- \`virglrenderer\` was fixed upstream in \`openembedded-core\` just over a week ago (commit 5ff5aac9ed9). However, it hasn't been backported to \`walnascar\` yet.
- \`mesa\` has not received a corresponding backport for this C11 threads header conflict in OE-core. 

Since these patches aren't yet available on the OpenEmbedded base layers Asteroid uses, shall I keep them here temporarily, or do you prefer I submit them to \`openembedded-core\` \`walnascar\` branch and wait for the Yocto backports to merge before fixing AsteroidOS?

beroset:
Great job on this, thanks!  Since `walnascar` is beyond EOL, upstream will probably not accept such patches.  I have also discovered that we also need a patch for `flex`: https://errors.yoctoproject.org/Errors/Details/853017/   Which watch are you building for?  I'm planning to try this on `catfish` today.

KonTy:
I bought TicWatch 3 ultra but I need to make a usb cable for it, so I'm working in a QEMU right now. My ultimate goal for hte first 2 weeks is to add a Power Manager where you can add Power profiles. Like in garmin where you can have profiles where you specify when sensors and radios are on and intervals for them. Comming from Enduro 3 I really want us to last for as long as we can squeeze from it. I was reading about 4100+ and thier RTOS processor. sucks we can't inject our own OS into it. it would be awesome to write our own RTOS. 

Are you ok if I fix QEMU images? right now everything is broken I can't click on things, I can't do swiping 90% of the time. I think it would help me speed up UI development if I can just boot an image on my machine.

PS do you have a usb cable for ticwatch that you can sell? I'm trying to make one following github post here

as you I wrote this health app and was looking for a first class citizen watch to pair it up with. But all the watches these days are encrypted and very closed source. Your OS is AWESOME and perfect. Please do not even stop working on it. We need more open source in this world.

 I would like to help you turn this watch into health research platform with sleep phase detection. I have some experience in that area where we can use some oxford AI models that are better then mainstream watches (garmin fitbit) to detect sleep phases. 

I would like to add workouts where you can chose data screen and power profiles, so you can determine which sensors are on for which workout.
Many things I want to add to the project. not much time :) 

**Let me know what your highest priorities are at this point can I help you get through these like tools, libs, compilers incompatibilities etc. ** so we can focus working on fun stuff

KonTy:
@beroset @FlorentRevest 
Do you guys hang out somewhere? Matrix/Discord or whatever? Can you tell me your plans for AsteroidOS? I want to help you bring it up to modern tools ASAP so we can focus on important things. Please let me know what I can do to help.

FlorentRevest:
Sure :) we're on Matrix https://asteroidos.org/contact/ this is where all discussions happen and would be the right place to get informed about project plans

MagneFire:
Re the nfcd fix, I think it's also fine to have the patch here. That way we aren't waiting for upstream to merge it.

KonTy:
> Re the nfcd fix, I think it's also fine to have the patch here. That way we aren't waiting for upstream to merge it.

@MagneFire @beroset 

Thank you guys, these changes are small, I know I'm not a meneber of hte team, but I vote for moving faster a little bit going forward we could submit patches in both places here and upstream which will allow us to go faster if you guys want.

PS I'm working on fixing QEMU emulator builds, what a drag, I've been at it for a week going one by one, It is a hard problem to solve. I think having a qorking emulator build will help UI development faster and will allow us to run simulations and fix memory leaks.

FlorentRevest:
Since no maintainer was able to reproduce this issue yet and it's likely to be fixed as part of regular updates, we won't be taking that tech debt into meta-asteroid for now. We can always reopen this if it becomes critical someday

PR #238, opened 2026-05-05
Add support for modern Android boot images

Aiden-Isik:
This pull request adds support for the (substantially more complex than before) modern Android device boot images.

Summary of changes:
* Added support for all current Android boot formats and partitions
* Added variables which can be set in machine/<$MACHINE>.conf to select which boot format and which boot partitions are used
* Added support for generating a blank but valid DTBO partition, as this is also required in the boot sequence of certain devices
* Renamed generated boot images to the names of the partitions they are flashed to.

On renaming the images specifically, I did discuss briefly in the Matrix room about this, and there wasn't really any opposition that I noticed, but if anyone does want this done differently feel free to let me know.

There is a corresponding pull request in the https://github.com/AsteroidOS/asteroidos.org repository (https://github.com/AsteroidOS/asteroidos.org/pull/496) to support these changes (only changes to make sure current devices don't break for now), however there also need to be changes made in the deploy script which moves the built images to asteroidos.org before this can really be merged. I was not able to find where this script is hosted, so if anyone could point me towards that it would be very helpful.


dodoradio:
hi! I've tested this PR with building a boot image for narwhal, and that's worked fine. I imagine more work needs to be done testing some other watches. @moWerk I think you should test this on some of the more obscure devices - probably tetra? But on a normal device like narwhal this causes no regressions.

I feel mixed about renaming all images to just partition name - I think it's useful to have the watch codename in there. The current `zImage-dtb-narwhal.fastboot` is excessive, perhaps a balance would be something like `boot-narwhal.img` or `narwhal.boot.img` or `asteroid-image-narwhal.boot.img` or `asteroid-narwhal.boot.img`

Aiden-Isik:
> I feel mixed about renaming all images to just partition name - I think it's useful to have the watch codename in there. The current `zImage-dtb-narwhal.fastboot` is excessive, perhaps a balance would be something like `boot-narwhal.img` or `narwhal.boot.img` or `asteroid-image-narwhal.boot.img` or `asteroid-narwhal.boot.img`

I agree that the codename would be useful. As discussed in the Matrix room, I've now renamed the images to follow the format ${DISTRO}-${MACHINE}-${IMAGE_NAME} (e.g. asteroid-lucky7-vendor_boot.img).

FlorentRevest:
Apologies for the time it took me to get to this... :) I had to edit our server's deploy script to deal with the new file naming.

It could still be temporarily broken but let's monitor https://release.asteroidos.org/nightlies/ in the next few days for evidences of breakages or that it works.

Could you update https://github.com/AsteroidOS/asteroidos.org/pull/496 too ?

PR #244, opened 2026-05-16
mce: ship explicit blanking defaults to fix display never blanking on Qt6

moWerk:
On fresh Qt6 installs, the display never auto-blanks after first boot under certain circumstances. One of them i could reproduce was booting up first time with 100% charge and usb connected on sparrow.

mcetool status shows Blank inhibit: stay-on and Tklock autoblank policy: disabled, meaning MCE correctly detects user inactivity but refuses to start the dim and blank sequence regardless.

Inspection of /var/lib/mce/builtin-gconf.values on first boot reveals inhibit_blank_mode=3 and tklock_blank_disable=1. The currently shipped file only contains display_brightness=5 with no overrides for these keys, so whatever values end up in the file at runtime are what MCE runs with.

Issuing mcetool --set-inhibit-mode=disabled and
mcetool --set-tklock-blank=enabled, which write inhibit_blank_mode=0 and tklock_blank_disable=0 to the same file, restores correct blanking behaviour and persists across reboots. Adding these as explicit shipped defaults achieves the same result from first boot without manual intervention.

Confirmed broken on fresh first boot on sparrow (MSM8226, kernel 3.10) and belugaxl (MSM8909W, kernel 4.9). Confirmed fixed on both after the equivalent mcetool commands.
Afaiu and could test, the zero values are the standard disabled/allow-blanking defaults for any Nemo MCE setup and do not affect devices where these keys were already correct.

### DIFF
--- a/recipes-nemomobile/mce/mce/builtin-gconf.values
+++ b/recipes-nemomobile/mce/mce/builtin-gconf.values
@@ -1,1 +1,3 @@
 /system/osso/dsm/display/display_brightness=5
+/system/osso/dsm/display/inhibit_blank_mode=0
+/system/osso/dsm/locks/tklock_blank_disable=0


MagneFire (commented):
Funny that this was never observed before.

Changes look good, but can you also apply these to the watch specific configurations?

Example: https://github.com/AsteroidOS/meta-smartwatch/blob/master/meta-catfish/recipes-nemomobile/mce/mce/builtin-gconf.values. This is just the first one I found, maybe there are other watches that use their own custom configuration.

FlorentRevest:
I'm surprised this is necessary when both `MCE_DEFAULT_TK_AUTO_BLANK_DISABLE` and `MCE_DEFAULT_BLANKING_INHIBIT_MODE` are already 0 https://github.com/sailfishos/mce/blob/f94a4251d9c34d8c4c0026e3a091bd01ee7015c8/tklock.h#L47 and https://github.com/sailfishos/mce/blob/f94a4251d9c34d8c4c0026e3a091bd01ee7015c8/modules/display.h#L438 

What I fear is that if we tactically fix those two without understanding the deeper issue, we might also have other invalid values

FlorentRevest:
Ah! This is probably due to https://github.com/AsteroidOS/asteroid-launcher/blob/0f676ef5e87e481a71fac527c0c72c63d607d179/src/firstrun.cpp#L47 not properly being reverted by https://github.com/AsteroidOS/asteroid-launcher/blob/0f676ef5e87e481a71fac527c0c72c63d607d179/src/firstrun.cpp#L54 ?

Maybe we somehow don't reach https://github.com/AsteroidOS/asteroid-launcher/blob/0f676ef5e87e481a71fac527c0c72c63d607d179/src/qml/firstrun/Tutorial.qml#L463 ?

PR #248, opened 2026-06-06
ngfd, usb-moded, bluebinder, distro, pulseaudio: add changes required by aurora

FlorentRevest:

PR #268, opened 2026-06-15
distro: resolve llvm-native and clang-native sysroot conflict

moWerk:
When meta-clang is present in the layer stack, `clang-native` and
oe-core's `llvm-native` both attempt to install the full LLVM C API
headers into native recipe sysroots. This causes `do_prepare_recipe_sysroot`
to abort with a duplicate file error, blocking Qt6 builds for any watch
that is not aurora.

The fix from PR #248 (hybris-gki-watch.inc) works correctly but is scoped
to GKI watches only. Moving it to `asteroid.conf` covers all watches
building with meta-clang and is a no-op when meta-clang is absent.

Note: once PR #248 merges, the same two lines should be removed from
`conf/machine/include/hybris-gki-watch.inc` to avoid redundancy.


FlorentRevest:
Thank you!

PR #273, opened 2026-06-17
timed: Don't grant cap_sys_time

FlorentRevest:
Nowadays we only change time via systemd calls (in both qml-asteroid and asteroid-btsyncd) so we no longer use timed's ability to change the current time.

In fact, it's actively hurting us because on aurora for example it is changing system time to a broken RTC which sets back time to 1970 and which breaks assumptions made by some Google libraries and restart loops some daemons in the LXC container that both break some functionalities and make the UI significantly slower.

### DIFF
--- a/recipes-nemomobile/timed/timed_git.bb
+++ b/recipes-nemomobile/timed/timed_git.bb
@@ -41,10 +41,9 @@ do_install:append() {
     ln -s /usr/share/zoneinfo/Etc/GMT ${D}/var/lib/timed/localtime
 
     cp ${UNPACKDIR}/timed.conf ${D}/etc/dbus-1/system.d/
-}
 
-pkg_postinst:${PN}() {
-    setcap cap_sys_time+ep $D/usr/bin/timed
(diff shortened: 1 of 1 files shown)


MagneFire:
> Nowadays we only change time via systemd calls (in both qml-asteroid and asteroid-btsyncd) so we no longer use timed's ability to change the current time.
> 
> In fact, it's actively hurting us because on aurora for example it is changing system time to a broken RTC which sets back time to 1970 and which breaks assumptions made by some Google libraries and restart loops some daemons in the LXC container that both break some functionalities and make the UI significantly slower.

In case you weren't aware this has been an issue for a while now on many watches, that's why we have https://github.com/AsteroidOS/meta-asteroid/blob/master/recipes-android/swclock-offset/swclock-offset_git.bb

FlorentRevest:
Yeah - it seems that timed wasn't bringing anything to the table since swclock-offset is the one that was actually restoring time correctly.

PR #279, opened 2026-06-28
ngfd: add a "notification" event that plays the notification sound

stianlaa:
asteroid-launcher currently plays notification/unmute sounds via QtMultimedia ⁠SoundEffect, which keeps a PulseAudio stream open for the app’s lifetime. This prevents the audio sink from suspending and causes constant standby battery drain.
The launcher will switch to ⁠ngf, which plays sounds with a short-lived GStreamer pipeline that shuts down after playback, allowing the sink to suspend again. To support this, this change adds a new sound-only ⁠"notification" event to ⁠ngfd (existing ⁠notif_* events remain for haptics/LED).

### DIFF
--- a/recipes-nemomobile/ngfd/ngfd/events.d/notification.ini
+++ b/recipes-nemomobile/ngfd/ngfd/events.d/notification.ini
@@ -1,0 +1,3 @@
+[notification]
+sound.filename = /usr/share/sounds/notification.wav
+sound.stream.media.role = notification


stianlaa:
Reopened: the review of AsteroidOS/asteroid-launcher#271 concluded the ngf path is the right fix after all - it solves the Qt6 chop (AsteroidOS/asteroid#342) as well as the standby drain, which the pure-Qt route could not. This event is the missing piece and should land before asteroid-launcher#270.

One open question from the description worth a maintainer opinion: this is a shared ngf event, so any other component emitting "notification" would also play the sound. Fine by design, or would you prefer a launcher-scoped event name?

FlorentRevest:
Fine by design

FlorentRevest (approved):

PR #280, opened 2026-07-02
Revert "usb-moded: Tolerate missing init_ffs"

dodoradio:
This reverts commit 73898c9993cd96c773d42837da8e7f3fc0dacfae. While this seems to be needed for aurora, it seems to also break USB adb on a bunch of post-pie watches
This breaks usb on aurora. https://github.com/AsteroidOS/meta-smartwatch/pull/318 may fix it by moving this change to be aurora-specific, but it's untested


dodoradio:
@FlorentRevest Thank you for the quick merge!

PR #282, opened 2026-07-06
Add support for the uniLoader second-stage bootloader

Aiden-Isik:
This pull request adds a package and a bbclass for the uniLoader second-stage bootloader. It is designed as a small shim which bundles the kernel, dtb, and initramfs inside itself, and provides a clean environment for booting Linux, free of the stock bootloader's interference (for example, adding hardcoded device tree overlays to the provided device tree).

This is mainly useful for ports using a close-to-mainline Linux kernel, although it could very well be possible to use it on downstream-based ports too if that were ever desired for some reason.

### DIFF
--- a/classes/uniloader.bbclass
+++ b/classes/uniloader.bbclass
@@ -1,0 +1,50 @@
+#
+# This class builds the built Linux kernel, DTB, and initramfs into uniLoader
+#
+
+DEPENDS:append = " uniloader"
+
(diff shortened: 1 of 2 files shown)


FlorentRevest:
Thank you! I know it took you a while but it's a good recipe :)

Aiden-Isik:
Thanks :D

PR #290, opened 2026-07-25
Drop DSME, libdsme & libiphb

FlorentRevest:
We were only using DSME for:

- _Device "state" management_: it was used by asteroid-launcher to detect shutdown (but that's now detected from systemd) and it is currently used by mce (but this PR makes it default to user mode which we probably never differed from anyway)
- _Setting up timers_: as far as I can tell (maybe needs more testing), on all our watches, it was using timerfds (which is a modern kernel infrastructure for that) exposed via a "iphb" socket which was usually used via a "libiphb" library. But by granting CAP_WAKE_ALARM to asteroid-launcher and its children processes, they can directly use timerfds instead and don't need to communicate using a library and a socket to a privileged system daemon

This PR reworks mce, nemo-keepaliva and asteroid-launcher to use timerfds and to drop all dependencies on libdsme, dsme and libiphb. This should make our dependencies tree a bit easier to reason about.


recipes-nemomobile/mce/mce_git.bb
```diff
@@ -12,13 +12,15 @@ SRC_URI = "gitsm://github.com/sailfishos/mce.git;protocol=https;branch=master \
     file://0006-Ambient-Mode-Exit-ambient-mode-when-touch-is-detecte.patch \
     file://0007-powerkey-Also-suspend-on-palm-reports.patch \
     file://builtin-gconf.values \
-    file://mce.service"
+    file://mce.service \
```
MagneFire, line 15: File order seems a bit weird here. Can you fix that?

recipes-nemomobile/nemo-keepalive/nemo-keepalive_git.bb
```diff
@@ -6,14 +6,15 @@ LIC_FILES_CHKSUM = "file://license.lgpl;md5=4fbd65380cdd255951079008b364516c"
 SRC_URI = "gitsm://github.com/sailfishos/nemo-keepalive.git;protocol=https;branch=master \
            file://0001-Disable-tests.patch \
            file://0002-Fix-build.patch \
+           file://0001-heartbeat-use-a-timerfd-directly.patch \
```
MagneFire, line 9: Numbering is wrong here

PR #292, opened 2026-07-25
underclock: add a single generic recipe for all watches

FlorentRevest:
Every watch shipped its own recipes-core/underclock recipe that was identical apart from a device-tuned script. Introduce one generic, machine-independent recipe here so the per-device copies in meta-smartwatch can be dropped.

The script guards every sysfs write behind an existence/writability check, so a single script adapts to whatever a given kernel exposes:

- offline every CPU core except cpu0 (adapts to the core count, so dual-core watches only drop cpu1)
- pick the best available CPU governor: schedutil (mainline since 4.7) when offered, else ondemand -- this reproduces each watch's previous choice from kernel age alone, with no per-device configuration
- pin the GPU to its lowest available frequency, read from kgsl-3d0/devfreq/available_frequencies (no hardcoded MHz)
- enable the lpm_levels / lpm_levels_legacy sleep and idle knobs
- prime MediaTek deep sleep via /dev/spm where present
- apply tetra's PowerHAL fixups (interactive governor + chown of the kona-fb / interactive tunables) only when the tetra-only kona-fb platform device exists

### DIFF
--- a/recipes-core/underclock/underclock/COPYING
+++ b/recipes-core/underclock/underclock/COPYING
@@ -1,0 +1,674 @@
+                    GNU GENERAL PUBLIC LICENSE
+                       Version 3, 29 June 2007
+
+ Copyright (C) 2007 Free Software Foundation, Inc. <http://fsf.org/>
+ Everyone is permitted to copy and distribute verbatim copies
+ of this license document, but changing it is not allowed.
+
+                            Preamble
+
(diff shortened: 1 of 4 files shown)


FlorentRevest:
This goes hand in hand with https://github.com/AsteroidOS/meta-smartwatch/pull/329

PR #293, opened 2026-07-25
libhybris: suppress modern-toolchain warnings for all machines

FlorentRevest:
The android-headers providers don't declare native_handle_clone and some HAL callbacks (e.g. gps_callback) cast between incompatible pointer types, so libhybris trips -Werror=implicit-function-declaration / int-conversion / incompatible-pointer-types with modern toolchains. This affects every machine, so suppress the warnings once here instead of per-device.


FlorentRevest:
This goes hand in hand with https://github.com/AsteroidOS/meta-smartwatch/pull/330

PR #294, opened 2026-07-25
Split the asteroid-launcher environment config into 3 files

FlorentRevest:
For now, the asteroid-launcher envvars are defined in /var/lib/environment/compositor/default.env and this file is overridden by all smartwatch layers.

This has caused some maintenance overhead because there are values which we would like to put in the environment of asteroid-launcher no matter the watch so those would have to be copied into each smartwatch layers. (that could be for features such as qtvirtualkeyboard and it's really awkward to modify smartwatch layers to make a new nistro feature work). Also, this has caused a lot of duplication, for example all hybris watches set a hwcomposer QPA which had to be copied 30 times when we could simply write it once in meta-asteroid based on whether the MACHINE inherits hybris-watch.

This change splits the environment into three files:

- 00-default.conf: which applies to all watches
- 10-platform.conf: which is either setting eglfs as QPA or hwcomposer, depending on whether the watch inherits hybrid-watch or not
- 20-device.conf: which sets additional watch-specific envvars only if necessary

---

Additionally, to reduce the number of watch specific changes required, it uses a new syntax for the evdevtouch plugin such that Qt auto-discovers touch screens and therefore we don't need to manually find the evdev file backing the touchscreen on every port. This may cause some surprises and would be good to test on a few watches.


FlorentRevest:
This goes hand in hand with https://github.com/AsteroidOS/meta-smartwatch/pull/331

MagneFire:
This should probably be checked but `minnow` uses the `fbdev` backend. This PR changes that to `hwcomposer`.

FlorentRevest:
Have you checked `meta-minnow/recipes-asteroid/asteroid-launcher/asteroid-launcher-configs/20-device.conf` in https://github.com/AsteroidOS/meta-smartwatch/pull/331 ? Variables from 10-platform.conf can be overridden

beroset:
Tracking state per watch

Platforms:
| Name             | Built | Tested | Notes          |
|------------------|-------|--------|----------------|
| anthias	| ✅ | ⚪ |  Don't own watch            |
| aurora	| ✅ | ⚪ |  compiles with [fix](https://github.com/AsteroidOS/asteroid/pull/357)      |
| bass	| ✅ | ⚪ |  Don't own watch            |
| beluga	| ✅ | ✅ |   tested by beroset   |
| catfish	| ✅ | ✅ |   tested by beroset           |
| dory	| ✅ | ✅ |  tested by beroset            |
| emulator	| ✅ | ✅ |  tested by beroset            |
| firefish	| ✅ | ⚪ |  Don't own watch            |
| harmony	| ✅ | ⚪ |  Don't own watch            |
| hoki	| ✅ | ✅ | tested by beroset             |
| inharmony	| ✅ | ⚪ | Don't own watch             |
| koi	| ✅ | ⚪ |  Don't own watch            |
| lenok	| ✅ | ⚪ | Don't own watch             |
| medaka	| ✅ | ⚪ |  Don't own watch            |
| minnow	| ✅ | ✅ | tested by MagneFire         |
| mooneye	| ✅ | ⚪ | Don't own watch             |
| narwhal	| ✅ | ⚪ | Don't own watch             |
| nemo	| ✅ | ✅ |  tested by beroset            |
| pike	| ✅ | ⚪ | Don't own watch             |
| ray	| ✅ | ⚪ | Don't own watch             |
| rinato	| ✅ | ⚪ | Don't own watch             |
| rubyfish	| ✅ | ⚪ | Don't own watch             |
| sawfish	| ✅ | ⚪ | Don't own watch             |
| skipjack	| ✅ | ✅ | tested by beroset             |
| smelt	| ✅ | ⚪ | Don't own watch             |
| sparrow	| ✅ | ✅ |  tested by beroset            |
| sparrow-mainline	| ⚪ | ⚪ |              |
| sprat	| ✅ | ⚪ |  Don't own watch            |
| sturgeon	| ✅ | ✅ |  tested by beroset            |
| swift	| ⚪ | ⚪ | Don't own watch              |
| tetra	| ⚪ | ⚪ | Don't own watch              |
| triggerfish	| ⚪ | ⚪ |  Don't own watch            |
| wren	| ✅ | ✅ |  tested by beroset            |

Legend:
- ✅: Tested, works as intended
- ⚪: Has not been tested
- 🟡: Tested, but has **existing** issues
- ❌: Tested, but has **new** issues

MagneFire (approved):

MagneFire:
Turns out, narwhal was broken. But now fixed via https://github.com/AsteroidOS/meta-smartwatch/pull/350

PR #297, opened 2026-07-26
Disable empty root login, root ssh login

dodoradio:
Yocto will set an invalid root password, disallowing any password login. Root login will also be disabled in sshd_config

For usecases where it's needed, root empty password can be re-enabled by running:
sudo passwd -d root

root ssh login can be reenabled with:
sudo sed -i 's/^[#[:space:]]*PermitRootLogin.*/PermitRootLogin yes/' /etc/ssh/sshd_config

### DIFF
--- a/classes/asteroid-image.bbclass
+++ b/classes/asteroid-image.bbclass
@@ -6,7 +6,7 @@ LICENSE = "GPL-2.0-only"
 inherit populate_sdk_qt6
 inherit asteroid-users
 
-IMAGE_FEATURES += "package-management empty-root-password allow-empty-password allow-root-login"
+IMAGE_FEATURES += "package-management allow-empty-password"
 
 IMAGE_INSTALL += " \
 kernel-modules base-files base-passwd systemd busybox iproute2 connman pam-plugin-loginuid bluez5 polkit polkit-group-rule-datetime sudo \


MagneFire:
Have you tested this with `adb` shell as well? 

Also, what's your reason for this change?

beroset:
Would `su -` would still work as the `ceres` user?

dodoradio:
I have tested this with adb shell. adb shell continues to give root access.
The reasoning behind this change has been discussed a lot, but to summarise it here:
- it's a basic security practice on every consumer linux distro
- it encourages better security practices for our users, a large proportion of whom are having their first experience with linux through asteroidos
- it has few drawbacks (the most meaningful ones are the small inconvenience to developers, for whom the revert commands above are provided)

I'm proposing this as part of an effort to improve security just the tiniest bit, so that we can then add a WiFi UI. The previous step was [adding sudo](https://github.com/AsteroidOS/meta-asteroid/commit/e5080b4d84e9acd7019fb0b9aaa3fda72750f000), the next step is to figure out some kind of UI for enabling/disabling remote login, through mechanisms such as disabling the SSH daemon, or adding a UI for setting a user password.

dodoradio:
> Would `su -` would still work as the `ceres` user?

no. sudo was added yesterday as a safer alternative

classes/asteroid-image.bbclass
```diff
@@ -6,7 +6,7 @@ LICENSE = "GPL-2.0-only"
 inherit populate_sdk_qt6
 inherit asteroid-users
 
-IMAGE_FEATURES += "package-management empty-root-password allow-empty-password allow-root-login"
```
FlorentRevest, line 9: I think I'd be ok with only removing "empty-root-password" from the default image. This should already address your concern that by default no one will be able to ssh into root or use su. Developers could just abandon this image entirely since it becomes painful to do any work on it and use asteroid-image-dev instead.

But I really don't see the point of removing allow-root-login. Until the user sets a root password no one can login anyway right ? And if they choose to set a root password, that means they want to log into it and in principle only they know the password so I don't see the security angle here.

Tangentially - I haven't tested sudo on a watch for now but how does it work right now, does it ask for a password or does it just grant access immediately because ceres doesn't have a password anyway ?

dodoradio: Ultimately, we don't want unaware users logging in as root, it encourages bad security practice. 
These two IMAGE_FEATURES disable security features for the convenience of developers. Removing one but not the other offers little benefit by itself. Developers would need to disable some of these security features on their own watches anyway, I don't think two commands vs. one is a significant difference. If they need root access, developers can build dev images or run two commands.

I also gotta say: For my dev work, I don't actually use root login for `devtool deploy` and similar operations (partly this is because I can't, as my build server is usually on a different network to my watch). I recognise the utility of the root account, I'm sure it makes iterations faster by automating deployment, and this is why I've documented the two commands that it takes to reenable it. 

Sudo doesn't currently ask for a password. When I filed this PR, I was looking at options for giving ceres a default password, or setting one through UI, which would prevent this behaviour. I'm currently looking at packagekit for app installs, I'll explore password options again later.

FlorentRevest: > Ultimately, we don't want unaware users logging in as root, it encourages bad security practice.

Sure. If users have to manually set a root password, this point is already addressed.

> These two IMAGE_FEATURES disable security features for the convenience of developers. Removing one but no the other offers little benefit by itself.

It... Does? Isn't it for example the same behavior as Debian? Or really anything but Ubuntu.

> Developers would need to disable some of these security features on their own watches anyway, I don't think two commands vs. one is a significant difference. If they need root access, developers can build dev images or run two commands.

I think it does make a difference. While I understand your justification for enforcing a root password, I don't understand your justification for disabling root login. This seems like vibe-based security theater.

dodoradio: all that `allow-root-login` does is run this command `sudo sed -i 's/^[#[:space:]]PermitRootLogin./PermitRootLogin yes/' /etc/ssh/sshd_config` at build time. 

I don't think there's any good justification for continuing to do that on non -dev images.

FlorentRevest: So while I find your argument for enforcing a root password technically grounded, I find that your argument for switching to an ubuntu style "root-login disabled over SSH by default" rather than a debian style "root-login over SSH enabled by default" (while requiring a password) boils down to "I just feel like it". From my perspective it's taking steps in the wrong direction to force users to run extra commands when they already have to go through extra hoops to enable any kind of root login anyway.

At this point, this no longer feels like a technical discussion but a totally vibe-based discussion and I have no interest in participating in those. If you have technical points to make about how this would genuinely make a difference in terms of security then I'm ready to hear them and change my mind but if this is just more "I feel like it" then I'll disengage from this PR review.

dodoradio: Yes. This is a debatable topic, the solutions have advantages and drawbacks. I prefer one solution, you prefer the other. That may seem 'vibe based'.
Most distros actually ship with  `PermitRootLogin prohibit-password`, I can't find any evidence that it's unique on debian or ubuntu. What I can confirm:
`PermitRootLogin prohibit-password` is the default on fedora
`PermitRootLogin prohibit-password` is also the default on yocto, unless it's overridden by `allow-root-login` (this is set at https://github.com/openembedded/openembedded-core/blob/master/meta/recipes-connectivity/openssh/openssh/sshd_config )

FlorentRevest: Thank you - now we can get unstuck from discussing preferences and I feel that this last paragraph brings a nuance that was missing from the framing before.

If Yocto's default is `prohibit-password`, doesn't it mean that after your PR, root login would still be allowed but only via public/private key pairs ? This is already a different storyline than disabling root login entirely as the description would make it sound. I'd need to think about it some more and again encourage other maintainers to chime in but this feels already a lot more acceptable to me.

beroset: I will be reviewing this today. One thing we will also need is something on the wiki to describe how it works, including with `devtool`.

dodoradio: I'm happy to write a wiki page, or leave it up to you if you'd prefer. 
I think we need to hit at least the following topics: (it might need to be multiple pages, or added to existing pages)
- the root account is 'disabled' by default
  - how to use sudo to avoid directly logging in as root
  - why you might want to log in as root
    - mainly, to use devtool deploy-target
  - how you can do it
    - reenabling root password login 
      - the two commands in this commit
    - adding an ssh key
      - what is an ssh key?

beroset: OK, I've done some pretty thorough testing, and I think we're ready to merge and that this is the right way to do it.  The two changes are each summarized separately:

1. remove `empty-root-password`

I think this one is pretty self-explanatory and I think we agree that now that wifi setup is available it's much safer to turn this off.  For access with a newly flashed watch, `adb` is now the default anyway and requires no password anyway for `adb shell`.

2. remove `allow-root-login`

Removing root login is what the option *says*, but here's the real world ramification.  While one can try to log in via ssh, and the watch will prompt for a password, the attempt will never succeed because of the invalid password.  However, one can still log in as `ceres`.  From there, `su -` will similarly prompt for password and then always give an "Authentication failure."  If one wants to do several things as root, logging in as `ceres` and then running `sudo su -` gives that effect and because the `ceres` has no password, it doesn't even prompt for a password.

One can copy the `.pub` ssh key from the development computer and append it to `/home/root/.ssh/authorized_keys`.  Log in once using `ssh root@watch` (and log out again) and thereafter `devtool deploy-target foo root@watch` works as always.  We can write all of this up in the `wiki` and I think it's a good step toward slightly better security without much user or developer pain.

beroset (approved):

MagneFire (approved):
All issues seem to be resolved. Thanks!

PR #299, opened 2026-07-27
Improve boot.img generation (less bbclasses - better recipe names)

FlorentRevest:


FlorentRevest:
This goes hand in hand with https://github.com/AsteroidOS/meta-smartwatch/pull/332

dodoradio:
this looks good. it's good to bring the initramfs recipe in-house, it was previously kinda hard to work with because it wasn't clear that it was being brought in from meta-smartphone.

PR #303, opened 2026-07-29
Systemd-conf: convert powerkey tweaks to drop-in

dodoradio:
Previously, the systemd-conf bbappend would install logind.conf to /etc/systemd/logind.conf, which would conflict with the systemd package, which installs to the same location. OPKG didn't notice this, but DNF does.
Use a drop-in instead.

Instead of doing this as a bbappend, it may make sense to turn this into its own recipe, on which MCE would depend. I've not done that yet to keep this change simple.


FlorentRevest:
From my perspective it would also be fair to do this as a mce bbappend rather than add another recipe just for that file

PR #305, opened 2026-07-29
android.inc: formalise qa skips

dodoradio:
I spotted issues here while trying to build with RPM/DNF packages, so I spent some time trying to untangle the QA skips. 

I'm not actually sure what the `ALLOW_EMPTY` did, but removing it hasn't broken ipk or rpm builds. This might need a review from @MagneFire , since he's the one who originally added that line. 

The 'fake' libstdc++.so RPROVIDE is a hack. The tarballs link to libstdc++, which we provide when we mount the system partition at boot time. 
Instead of adding the RPROVIDE, this could be worked around using `INSANE_SKIP:android-system += "file-rdeps"`. however, DNF does its own per-file dependency check, and there doesn't seem to be a way to skip that. 
Since it's just one vendor-provided file that these tarballs depend on, I feel this hack isn't too egregious.

Watches that package their own system directory (basically everything older than oreo) also need to be build-tested with RPM, though they're likely to have other quirks.

### DIFF
--- a/recipes-android/android/android.inc
+++ b/recipes-android/android/android.inc
@@ -24,12 +24,10 @@ do_install() {
     rm ${D}${includedir}/android/android-headers.pc
 }
 
-# Precompiled libraries, INSANE_SKIP all.
-do_package_qa() {
-}
-
-# Allow empty main package since the -dev package depends on it
-ALLOW_EMPTY:${PN} = "1"
+# We don't ship libstdc++.so as part of these tarballs. 
+# This library exists on these watches' system partitions, which we mount during boot.
+# Convince yocto that we already have this file.
+RPROVIDES:android-system += "libstdc++.so"
 
 PACKAGES =+ "android-system android-headers"
 FILES:android-system = "/usr"


FlorentRevest (approved):
Looks good to me but the lines that you dropped look like the kind of things that break on some random watches so we'd have to be careful that all watches build after that

dodoradio:
ok, I'm gonna make this a draft PR until I verify that all the watches still build. I'm quite sure these changes should just affect packaging, not builds or runtime

FlorentRevest:
do_package_qa is definitely treated at build time, I think the block that was there would basically disable all QA checks done by bitbake so it seems likely to me that things might start to report warnings on various watches, but maybe I'm wrong. 

Building for all watches is a pain in the butt (at least on my laptop I keep running out of space) so if you're not in a position where you can do that, I'm ok with merging it and then reacting to our nightlies builds maybe breaking. 

The reason why I did not submit is because you asked for a review from MagneFire

dodoradio:
I tried building every watch, everything succeeded, apart from:
- aurora. This watch seems to rebuild every package, so I stopped it because it was taking too long
- emulator. Different architecture, would need to rebuild everything, and it doesn't touch this android package anyway
- harmony. I ran into a patch issue with the launcher recipe? but I think the android recipe already succeeded by that point.
- minnow. This has a different armv7 tune, so it rebuilds everything, I also stopped it because it was taking too long.
- rinato. This is a mainline watch, it shouldn't touch this android package
- rubyfish. I ran into some issue with building the kernel, I couldn't get past that
- sparrow-mainline. Another mainline watch, shouldn't touch this android package

It's not an amazing success rate, but I don't think anything broke _because_ of the android package, so I think this is fine to merge. I'm undrafting this.

PR #311, opened 2026-08-03
Stop providing QT_IM_MODULE to QtWayland clients

FlorentRevest:
Since Qt6, QtWayland ignores a client-side
QT_IM_MODULE=qtvirtualkeyboard and complains in the logs every time an app starts with "qtvirtualkeyboard currently is not supported at client-side, use QT_IM_MODULES=qtvirtualkeyboard at compositor-side" We should have asteroid-launcher provide the keyboard in a unified manner using qtvirtualkeyboard, asteroid-virtualkeyboard and QT_VIRTUALKEYBOARD_STYLE=watch and let clients reach it over qt_text_input_method_v1:

For now, drop a horrible qtbase patch that defaulted an empty QT_IM_MODULE to qtvirtualkeyboard. It forced the client-side path even after the env var was removed, keeping the keyboard from routing to the compositor. Also stop setting QT_IM_MODULE=qtvirtualkeyboard in mapplauncherd booster's environments so that apps no longer have this in their environment


MagneFire:
This is not Qt6 unique, this message was also shown on Qt5.

MagneFire:
Can you confirm that `asteroid-launcher` actually provides the keyboard in the unified manner? Or is the keyboard entirely broken after this change?

EDIT: This seems to somewhat depend on https://github.com/AsteroidOS/meta-asteroid/pull/310?

PR #312, opened 2026-08-03
Remove stale bbappends and variables

FlorentRevest:


recipes-connectivity/openssh/openssh_%.bbappend
```diff
@@ -1 +0,0 @@
-PACKAGECONFIG:remove = "rng-tools"
```
MagneFire, line 1: Was surprised that this isn't needed anymore as this was used to solve high CPU usage on older kernels (https://github.com/AsteroidOS/meta-asteroid/commit/98c6df0e402cc17147c7660b66d95f552541cee5). But it seems that upstream removed it entirely (https://github.com/openembedded/openembedded-core/commit/868dfb46d96a27ec9041cb902fb769330277257d)

MagneFire (approved):

MagneFire:
Lovely cleanup, thanks!

PR #213, opened 2026-01-14
move most recipes to whinlatter

beroset:
This moves most recipes to whinlatter, as per https://github.com/AsteroidOS/asteroid/issues/322 but the there are still a few recipes that don't yet build.

PR #313, opened 2026-08-05
Move Yocto from Walnascar to Yocto Whinlatter

MagneFire:
Moves Yocto from Walnascar to Whinlatter.

This supersedes: https://github.com/AsteroidOS/meta-asteroid/pull/213


beroset:
Merged as #321

PR #315, opened 2026-08-08
adb: Use mincrypt & remove host tools from initramfs

MagneFire:
Currently OpenSSL is installed because adbd requires it. OpenSSL is rather large at 2.88 MB. This makes the boot image size come closer to the 10 MB limit used on `smelt`. Switch to libmincrypt to get rid of this dependency.

Additionally, remove the host side tools (adb, fastboot) from the initramfs image.


beroset:
Examined and tested.  Thanks!

PR #316, opened 2026-08-08
qt6-qpa-hwcomposer-plugin: Make v10 updates request-driven and vsync-paced

MagneFire:
`HwComposerBackend_v10` has no `requestUpdate()` override, so QtQuick falls back to its
generic 5 ms update timer and re-renders the compositor scene continuously — even when
nothing changed and even when the display is off. On minnow
that held the render thread near 90% CPU on a blanked screen.

It also blocks suspend: on blank, mce calls the compositor's
`org.nemomobile.compositor.setUpdatesEnabled(false)` and waits for rendering to stop before
releasing the `mce_display_on` wakelock. The v10 compositor never stops, so mce holds the
wakelock forever and the SoC never enters deep sleep. `v11`/`v20` already deliver updates on
demand; this brings `v10` in line.


MagneFire:
PR for the fallback patch is here: https://github.com/mer-hybris/qt5-qpa-hwcomposer-plugin/pull/106

PR #320, opened 2026-08-27
distro: fix package feed architecture names

monperrus:
Correct two entries in `PACKAGE_FEED_ARCHS` so generated opkg source URLs match the published nightly feed directories:

- `armv7athf-neon` → `armv7at2hf-neon`
- `sparrow-mainline` → `sparrow_mainline`

Both corrected `Packages.gz` URLs return HTTP 200. On a sawfish nightly image, the previous names caused `opkg update` to abort; with these replacements it completes successfully.


MagneFire (approved):
Good catch! Thanks for the fix 👍

PR #322, opened 2026-09-03
ngfd: Play notification sound via x-maemo media role

### LINKED ISSUE #296: Volume Mute Is Broken In Nightly Builds
there is an issue on Beluga where the mute button doesn't work and i think the volume slider is broken as well, when setting volume to 0% it sounds the same as 100% and 50% to my ears.

MagneFire:
AsteroidOS volume control always sets the mainvolume. mainvolume is mapped to the x-maemo media role.
This role is also used for any streams that don't set a role.

Recently, notification sound was moved away from QtMultimedia to ngfd. This resulted in notifications always sounding at 100% regardless of volume level set in the quickpanel or settings.

Using the x-maemo media role fixes this issue.

This fixes https://github.com/AsteroidOS/asteroid-launcher/issues/296.

### DIFF
--- a/recipes-nemomobile/ngfd/ngfd/events.d/notification.ini
+++ b/recipes-nemomobile/ngfd/ngfd/events.d/notification.ini
@@ -1,3 +1,3 @@
 [notification]
 sound.filename = /usr/share/sounds/notification.wav
-sound.stream.media.role = notification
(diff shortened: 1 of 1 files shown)


beroset (approved):
Tested and confirmed on both beluga and catfish

